PR #9, opened 2016-03-15
Add reuse_inactive_volume option to bigip_software module

festango:
Add an option to bigip_software module to dynamically select one of existing inactive volumes (or create new one if no inactive volume in the system). The motivation of the change is to allow playbook to install software without specifying a volume name to which software is installed. If you have a playbook to install a new software into multiple bigip nodes at once, you may not be able to explicitly specify a volume name upfront. 

This module will give users an option to automatically select install volume from one of inactive volumes.

**Change**
- Add `reuse_inactive_volume` option
```
  reuse_inactive_volume:
    description:
      - Automatically chooses the first inactive volume in alphanumeric order. If there
        is no inactive volume, new volume with incremented volume name will be created.
        For example, if HD1.1 is currently active and no other volume exists, then the 
        module will create HD1.2 and install the software. If volume name does not end with
        numeric character, then add '.1' to the current active volume name. When C(volume) 
        is specified, this option will be ignored.
    required: false
```

- Change behaviour of `force` option to force install software into a new volume even if the system is currently running the software user is trying to install.

```
  force:
    description:
      - If C(yes) will upload the file every time and replace the file on the
        device. If C(no), the file will only be uploaded if it does not already
        exist. Generally should be C(yes) only in cases where you have reason
        to believe that the image was corrupted during upload.
      - If C(yes) with C(reuse_inactive_volume) is specified and C(volume) is 
        not specified, Software will be installed / activated regardless of current
        running version to a new or an existing volume.
    required: false
    default: no
    choices:
      - yes
      - no
```
**ToDo**
-  It may be good to add a `prefix` option to the volume name so that the playbook does not accidentally remove important volume
-  It may be good to add feature to specify a list of volume names in which Ansible can remove it and install new software.

### DIFF
--- a/library/bigip_software.py
+++ b/library/bigip_software.py
@@ -36,6 +36,9 @@ options:
         device. If C(no), the file will only be uploaded if it does not already
         exist. Generally should be C(yes) only in cases where you have reason
         to believe that the image was corrupted during upload.
+      - If C(yes) with C(reuse_inactive_volume) is specified and C(volume) is 
+        not specified, Software will be installed / activated regardless of current
+        running version to a new or an existing volume.
     required: false
     default: no
     choices:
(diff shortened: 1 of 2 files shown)


caphrim007:
lgtm

PR #11, opened 2016-04-20
Correct bigip_device_dns example

buzzsurfr:
Quick add to correct the bigip_device_dns example included (it was a copy of the sysdb example).


caphrim007:
merging. thanks!

PR #18, opened 2016-06-13
Request to add bigip_ssl_certificate module

waffie1:
Requesting the addition of my bigip_ssl_certificate module.  This module will perform basic add/remove of an ssl certificate/key pair to a BigIP.

Module supports both iControl via bigsuds and REST via f5-sdk.   I believe REST requires BigIP v12 to support the file upload process(without having to manually create /var/config/rest/download/tmp) which is why bigsuds method has been included.

Currently only PEM format is supported.  Will add support for PKCS12 if/when f5-sdk adds support, or as time allows to implement local conversion to PEM first.

Script will currently exit unchanged if certificate name exists.  This was done temporarily until functionality is added to compare existing, modify when needed and exit unchanged when certs are identical.  Given the difficulties with replacing cert/key pairs, the usefulness of this is somewhat limited.


library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+# TODO: pkcs12 support when f5-sdk adds it, or convert locally to pem?
+#       Would be nice if f5-sdk file upload could take a string like soap
+#       cert/key import, then we could convert in memory and import
```
caphrim007: I will speak with Za about this, as I suspect it's within the realm of possibility to add

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+else:
+    bigsuds_found = True
+
+try:
+    from f5.bigip import ManagementRoot
+    from icontrol.session import iControlUnexpectedHTTPError
+except:
+    f5sdk_found = False
+else:
+    f5sdk_found = True
```
caphrim007: This can be changed to ``HAS_F5SDK``

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+'''
+
+
+try:
+    from OpenSSL import crypto
+
+except:
+    found_OpenSSL = False
+else:
+    found_OpenSSL = True
```
caphrim007: This can be moved in to the `try` block and remove the `else`.

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+def pkcs12_to_pem(pkcs12data, password=''):
+    pkcs12 = crypto.load_pkcs12(pkcs12data, password)
+    key = crypto.dump_privatekey(crypto.FILETYPE_PEM, pkcs12.get_privatekey())
+    cert = crypto.dump_certificate(crypto.FILETYPE_PEM,
+                                   pkcs12.get_certificate())
+    return (cert, key)
+
+
+class F5ModuleError(Exception):
+    pass
```
caphrim007: This is now a part of Ansible, so it can be removed

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+class BigIPCommon:
+    def __init__(self, hostname, username, password, verify=True):
+        self.hostname = hostname
+        self.username = username
+        self.password = password
+        self.verify = verify
+        self.api = self.connection(hostname, username, password, verify)
+
+
+class BigIPiControl(BigIPCommon):
```
caphrim007: If this module can work entirely with REST, I recommend we remove the SOAP functionality in the interest of driving REST adoption.

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+            cert_pem_file=dict(type='str'),
+            key_pem_file=dict(type='str'),
+            pkcs12_file=dict(type='str'),
+            pkcs12_password=dict(type='str', no_log=True),
+            validate_cert=dict(type='bool', default=True),
+            # overwrite=dict(type='bool', default=False)
+        )
+    )
+    result = {'changed': False}
+    if module.params['connection'] == 'icontrol':
```
caphrim007: If we can do without SOAP, then you can eliminate the entire statement here and just use the REST call

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+#       Would be nice if f5-sdk file upload could take a string like soap
+#       cert/key import, then we could convert in memory and import
+# TODO: Exception Handling
+
+from ansible.module_utils.basic import *
+
+DOCUMENTATION = '''
+module: bigip_ssl_certificate
+short_description: Import/Delete certificates from BIG-IP
+description:
```
caphrim007: Please add a `version_added: "2.2"` here

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+# TODO: pkcs12 support when f5-sdk adds it, or convert locally to pem?
+#       Would be nice if f5-sdk file upload could take a string like soap
+#       cert/key import, then we could convert in memory and import
+# TODO: Exception Handling
+
+from ansible.module_utils.basic import *
```
caphrim007: This can be moved to right before the `if __name__ == '__main__':` below

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+        choices: [true. false]
+notes:
+    - bigsuds required for icontrol connection
+    - f5-sdk version 0.1.7 required for rest connection
+    - PKCS12 will be supported when f5-sdk adds support.
+requirements:
+    - bigsuds
+    - f5-sdk
+author:
+    - Kevin Coming <kevcom@gmail.com>
```
caphrim007: You can just include your name and Github handle here

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+DOCUMENTATION = '''
+module: bigip_ssl_certificate
+short_description: Import/Delete certificates from BIG-IP
+description:
+    - This module will import/delete SSL certificates on BIG-IP LTM
+      systems from a certificate file on the filesystem where the playbook
+      is run from.  Currently this only supports certificates to be used
+      within SSL Profiles.  PEM format with certificate and key in seperate
+      files is icurrently the only supported format.
+options:
```
caphrim007: The `DOCUMENTATION` and related blocks at the tops of these files can be spaced with 2 spaces.

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+            - Currently unused
+        required: False
+        default: none
+    validata_cert:
+        description:
+            - Validate the certificate on the remote Big-IP is valid.
+            - Currently f5-sdk appears to hardcode this to False
+        required: False
+        default: True
+        choices: [true. false]
```
caphrim007: this should be a comma instead of a decimal

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+        description:
+            - Current unused
+        required: False
+        default: none
+    pkcs12_password:
+        description:
+            - Currently unused
+        required: False
+        default: none
+    validata_cert:
```
caphrim007: Spelling correction here; "validate"

library/bigip_ssl_certificate.py
waffie1: I've read the file upload does not reliably work until LTM v12.
Specifically, the /var/config/rest/download/tmp directory keeps deleting
itself(at an interval I haven't figured out yet) and needs to be manually
created.  I left the SOAP calls in place for people still on v11.  If that
is not desirable, it can be removed.

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,449 @@
...
+        except F5ModuleError as e:
+            module.fail_json(msg="F5ModuleError: %s" % e)
+        except ConnectioError as e:
+            module.fail_json(msg="F5ModuleError: %s" % e)
+        else:
+            if r is True:
+                result['changed'] = True
+
+    module.exit_json(**result)
+
```
caphrim007: You'll want to include the F5 module utils here.

`from ansible.module_utils.f5 import *`

caphrim007:
Thanks for the PR @waffie1 ! If you would be so kind as to address the comments above, I'll have another go at review.

library/bigip_ssl_certificate.py
caphrim007: Lemme ask the REST framework developers about that deletion. I'm fine with this being a 12 or greater module. I'm treating these modules (generally) as an incentive to upgrade. So I'm deliberately being picky about providing SOAP support.

If there is an issue with the deleting thing, you can also try performing the operations inside of a Transaction.

library/bigip_ssl_certificate.py
waffie1: inclusion of tm.sys.crypto.pkcs12 in the sdk would probably be the most
useful here.

My original through here was to convert PKCS12 to PEM in memory with
OpenSSL and then upload without having to write out a file.  After
discovering the native REST api has pkcs12 functions, this is completely
unnecessary if the f5-sdk exposes that api.

caphrim007: I'm also a member of the team that devs that, so I will investigate its inclusion. Our release cycle on the sdk is every 2 weeks.

waffie1:
I believe all outstanding issues have been resolved.   Please take another look.

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,322 @@
...
+  validate_cert:
+    description:
+      - Validate the certificate on the remote Big-IP is valid.
+        Currently f5-sdk appears to hardcode this to True
+        And thus this does nothing
+    required: False
+    default: True
+    choices: [true, false]
+notes:
+    - bigsuds required for icontrol connection
```
caphrim007: this can be removed

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,322 @@
...
+
+    - name: "Delete Certificate"
+      local_action:
+        name: certificate-name
+        module: bigip_ssl_certificate
+        server: bigip-addr
+        username: username
+        password: password
+        state: absent
+        connection: rest
```
caphrim007: `connection` param not needed

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,322 @@
...
+'''
+EXAMPLES = '''
+    - name: "Import PEM Certificate"
+      local_action:
+        name: certificate-name
+        module: bigip_ssl_certificate
+        server: bigip-addr
+        username: username
+        password: password
+        connection: rest
```
caphrim007: `connection` can be removed

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,322 @@
...
+    default: null
+  name:
+    description:
+      - SSL Certificate Name.  This is the cert/key pair name used
+        when importing a certificate/key into the F5.  It also
+        determines the filenames of the objects on the LTM
+        (:Partition:name.cer_11111_1 and :Partition_name.key_11111_1)
+    required: True
+    default: none
+  connection:
```
caphrim007: `connection` can be removed

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,322 @@
...
+        filename = os.path.join('/var/config/rest/downloads', keyfilename)
+        params = {'from-local-file': filename, 'name': name}
+        self.api.tm.sys.crypto.keys.exec_cmd('install', **params)
+
+    def upload(self, filename):
+        self.api.shared.file_transfer.uploads.upload_file(filename)
+
+
+def main():
+    module = AnsibleModule(
```
caphrim007: With the inclusion of `from ansible.module_utils.f5 import *`, this argument spec can be simplified.

If you take a look here

https://github.com/F5Networks/f5-ansible/blob/master/library/bigip_device_sshd.py#L335

from line 335 to line 360, you can use the same workflow to minimize this block.

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,322 @@
...
+        argument_spec=dict(
+            state=dict(type='str', default='present',
+                       choices=['absent', 'present']),
+            user=dict(type='str', required=True, aliases=['username']),
+            password=dict(type='str', required=True, no_log=True),
+            server=dict(type='str', required=True),
+            name=dict(type='str', required=True),
+            partition=dict(type='str', default='Common'),
+            connection=dict(type='str', default='rest',
```
caphrim007: `connection` not needed

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,322 @@
...
+            validate_cert=dict(type='bool', default=True),
+            # overwrite=dict(type='bool', default=False)
+        )
+    )
+    result = {'changed': False}
+    if module.params['connection'] == 'rest':
+        if not HAS_F5SDK:
+            module.fail_json(msg='connection type rest requires '
+                             'f5-common-python')
+        api = BigIPRest(module.params['server'],
```
caphrim007: if you take a look at line 355 here

https://github.com/F5Networks/f5-ansible/blob/master/library/bigip_device_sshd.py#L355

it'd be fantastic if you would follow a similar approach. That approach being to keep the execution bits to a minimum and to wrap the logic and processing inside of a `flush` method of an object named after the module. In this case, the class name would be `BigIpSslCertificate`.

caphrim007:
@waffie1 reviewed again; comments noted. Thanks for the work so far!

library/bigip_ssl_certificate.py
```diff
@@ -0,0 +1,322 @@
...
+            else:
+                if r:
+                    result['changed'] = True
+    elif module.params['state'] == 'absent':
+        # TODO Delete cert and key separately???
+        try:
+            r = api.delete(module.params['name'])
+        except F5ModuleError as e:
+            module.fail_json(msg="F5ModuleError: %s" % e)
+        except ConnectioError as e:
```
caphrim007: I believe you're missing an "n" here

library/bigip_ssl_certificate.py
waffie1: A question on the flush method approach.  Do we want as much logic as
possible there, or just logic specific to the tasks of interacting with the
F5?

Example for this module, the cert/key filenames are only required if state
is present.  If state is absent, they are not needed, so I did not make
them required in the argument_spec.  Should this logic still be inside of
the class, or should this be handled in main and exit when something is
missing, before we even call the class?

Kevin

caphrim007: the way I currently do it is to embed everything inside of the flush() method and make the main execution are short as I can.

waffie1:
Please review the latest version, hopefully I have addressed everything.  Thanks for your patience, I'm new to this and appreciate the guidance you are providing.

caphrim007:
Thanks for all the work @waffie1 !

Next, would you be so kind as to add the tests for it as well?

https://f5-ansible.readthedocs.io/en/latest/development/writing-a-module.html#testing

You can reference `bigip_device_sshd` if you need a working example. I use the tests moving forward to guarantee the working-ness of a module. You can include any certificate files inside the files/ directory of the role. For certs, including self-signed examples is fine. The valid lifetime for them can be set to something far in the future (like 9999 days or something).

If you have any special steps for creating items in the files or templates directories, you can use the README.md pattern I describe in the url above.

caphrim007:
@waffie1 this is really great and impressive stuff. I'll have a look-see.

caphrim007:
@waffie1 I'm going to let it bake in rotation here and I'll point some of my colleagues at it. Then we'll begin the upstream process and I'll include you in that

PR #20, opened 2016-07-14
enable route advertisement

mgusek:
This patch enables route advertisement for virtual server's ip.

### DIFF
--- a/library/bigip_virtual_server.py
+++ b/library/bigip_virtual_server.py
@@ -128,6 +128,11 @@ options:
       - Default Profile which manages the session persistence
     required: false
     default: None
+  route_advertisement_state:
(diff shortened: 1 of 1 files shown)


caphrim007:
@mgusek will merge, but it appears that this api call requires the address/destination to work correctly. So I will add a test for that.

Can you post the ansible task you were using to run this just so I can verify what I said above?

caphrim007:
@mgusek when you have time, post a task example for this, otherwise, I have some in the tests folder that you can try.

I will upstream this change

mgusek:
@caphrim007 as i can see, you already implement a test for route advertisements. Do you need more information from me ?

Issue #22, opened 2016-08-15
bigip_node.yml error


taka379sy:
bigip_node.yml will result in an error .
```
An exception occurred during task execution. The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_LUv4Q7/ansible_module_bigip_node.py", line 398, in <module>
    main()
  File "/tmp/ansible_LUv4Q7/ansible_module_bigip_node.py", line 295, in main
    (server, user, password, state, partition, validate_certs) = f5_parse_arguments(module)
ValueError: too many values to unpack

fatal: [bigip1 -> 127.0.0.1]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_name": "bigip_node"}, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_LUv4Q7/ansible_module_bigip_node.py\", line 398, in <module>\n    main()\n  File \"/tmp/ansible_LUv4Q7/ansible_module_bigip_node.py\", line 295, in main\n    (server, user, password, state, partition, validate_certs) = f5_parse_arguments(module)\nValueError: too many values to unpack\n", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
```

Change before
```
    (server, user, password, state, partition, validate_certs ) = f5_parse_arguments(module)
```
After the change
```
    (server, user, password, state, partition, validate_certs, server_port) = f5_parse_arguments(module)
```
After the correction , now OK.

caphrim007:
This error was due to a lack of merging in upstream on my part. The fix was made upstream and I had yet to bring said fix into the repo here

PR #23, opened 2016-08-15
issues #22

taka379sy:
issue "bigip_node.yml error  #22"
It was corrected.
Pull Request is sent.

### DIFF
--- a/library/bigip_node.py
+++ b/library/bigip_node.py
@@ -292,7 +292,7 @@ def main():
         supports_check_mode=True
     )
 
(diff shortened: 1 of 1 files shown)


caphrim007:
hey @taka379sy I think you've caught me red handed! It looks like this change is merged in upstream here

https://github.com/ansible/ansible-modules-extras/blob/devel/network/f5/bigip_node.py#L355

I have yet to pull those changes into the f5-ansible repo here. I will do that right now. Thanks for catching my mistake and for the contribution!

PR #31, opened 2016-09-27
Enabled Route Domain Functionality

manuadoor:
I have enabled the optional functionality of creating self IP for a specific route domain. Please review and consider approving the pull request.

- Manu Pillai 

Example: 

 - name: Create Self IP with a Route Domain
   bigip_selfip:
     server: "lb.mydomain.com"
     user: "admin"
     password: "secret"
     validate_certs: "no"
     name: "self1"
     address: "10.10.10.10"
     netmask: "255.255.255.0"
     vlan: "vlan1"
     rd: "10"
     allow_service: "default"
   delegate_to: localhost

### DIFF
--- a/library/bigip_selfip.py
+++ b/library/bigip_selfip.py
@@ -72,6 +72,7 @@ requirements:
   - f5-sdk
 author:
   - Tim Rupp (@caphrim007)
+  - Manu Pillai (@manuadoor)
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


caphrim007:
@manuadoor so to make this work we only need to append the route domain to the ip? is that correct?

manuadoor:
Thats correct.. we will pass an optional parameter "rd" for route domain id and self IP will be created for that rd.

caphrim007:
ok, gotcha. this PR needs to be rebased to take into account the recent traffic class additions. can you rebase from the current master branch?

manuadoor:
Alright, I will take a look at it soon

manuadoor:
I just checked, you are talking about the last commit (https://github.com/F5Networks/f5-ansible/pull/31/commits/114a082d93b59e2490a5a8d67c247e6a2f51274e) 6 days ago ? if yes, it is based on that commit.

caphrim007:
yes, but PR's need to not show those changes that I made, they must only show the changes you made. The PR you have here shows both my changes as well as your changes.

PR #70, opened 2017-02-06
Prevent Duplicate command detected

rklomp:
When multiple commands are given the disable_pager is duplicate and prevents all other commands from executing.


caphrim007:
@rklomp the bigip_command module has had a significant refactor done to it. I'm going to close this because this particular patch no longer applies, can you try using the new module to see if it meets your needs?

Issue #85, opened 2017-03-01
bigip_user fails on creating or deleting of user account.


slavkv:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME
bigip_user

##### ANSIBLE VERSION

```
ansible 2.2.1.0
  config file =
  configured module search path = Default w/o overrides
```

##### PYTHON VERSION

```
Python 2.7.13
```

##### BIGIP VERSION
```
Sys::Version
Main Package
  Product  BIG-IP
  Version  11.5.1
  Build    8.0.175
  Edition  Hotfix HF8
  Date     Mon Feb  9 00:49:11 PST 2015
```

##### CONFIGURATION
host_key_checking = False
ask_pass=False

##### OS / ENVIRONMENT
OSX 10.11.6

##### SUMMARY
Module 'bigip_user' fails when I try to create or delete user. I am using example from documentation.
Modules 'bigip_command' and 'bigip_user_facts' work fine.

##### STEPS TO REPRODUCE

```
  - name: Add user accounts
    bigip_user:
      server: "{{ device_hostname | default(inventory_hostname) }}"
      user: "{{ device_user }}"
      password: "{{ device_pass }}"
      validate_certs: false
      username_credential: "testuser5"
      password_credential: "adasdasdafsafasdf"
      partition_access: "all:admin"
      full_name: "John Doe"
      shell: "bash"
      state: "present"
    delegate_to: localhost
```

##### EXPECTED RESULTS

A user "testuser5" is created.

##### ACTUAL RESULTS

```
fatal: [myf5box -> localhost]: FAILED! => {
    "changed": false,
    "failed": true,
    "invocation": {
        "module_name": "bigip_user"
    },
[...]

PR #105, opened 2017-03-24
Fixes #29 #85

### LINKED ISSUE #29: bigip_user produced unhandled Python exception AdminRoleNoModifyError
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

F5 bigip (bigip_user)
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### PYTHON VERSION

```
Python 2.7.12
```
##### F5 BIGIP LTM VERSION

```
Sys::Version
Main Package
  Product     BIG-IP
  Version     12.1.1
  Build       0.0.184
  Edition     Final
  Date        Thu Aug 11 17:09:01 PDT 2016
```
##### CONFIGURATION

```
retry_files_enabled = False
host_key_checking=False
```
##### OS / ENVIRONMENT

Alpine 3.4 (Docker Container)
##### SUMMARY

When attempting to modify the root user's password using `bigip_user`, an unhandled Python exception is thrown with the error `NameError: global name 'AdminRoleNoModifyError' is not defined`.

_NOTE: after running into this error, I noticed in the documentation that this module is only supported on BIGIP versions 13 and greater. Whether or not it is supported on the version of software I'm using, a reasonable error should be produced instead of an unhandled exception._
##### STEPS TO REPRODUCE

Create a task that uses the `bigip_user` to attempt to change the password of the root user. Additionally, the BIGIP software version being tested against may need to be lower than 13.x to produce this error.

```
- name: Set root user password
  bigip_user:
    password: "{{ bigip_password }}"
    server: "{{ inventory_hostname }}"
    user: "{{ bigip_username }}"
    username_credential: "root"
    password_credential: "mysecret"
    validate_certs: "{{ validate_certs }}"
```
##### EXPECTED RESULTS

It should handle the `AdminRoleNoModifyError` exception and return a reasonable error message indicating the actual problem.
##### ACTUAL RESULTS

The Ansible task fails with the following error: `NameError: global name 'AdminRoleNoModifyError' is not defined`.

```
TASK [Set root user password] **************************************************
task path: /playbooks/ltm_bootstrap.yaml:9
Using module file /playbooks/library/bigip_user.py
<lbl12.example.com> ESTABLISH LOCAL CONNECTION FOR USER: root
<lbl12.example.com> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1474561274.87-25040398694960 `" && echo ansible-tmp-1474561274.87-25040398694960="` echo $HOME/.ansible/tmp/ansible-tmp-1474561274.87-25040398694960 `" ) && sleep 0'
<lbl12.example.com> PUT /tmp/tmpZGCFpX TO /root/.ansible/tmp/ansible-tmp-1474561274.87-25040398694960/bigip_user.py
<lbl12.example.com> EXEC /bin/sh -c 'chmod u+x /root/.ansible/tmp/ansible-tmp-1474561274.87-25040398694960/ /root/.ansible/tmp/ansible-tmp-1474561274.87-25040398694960/bigip_user.py && sleep 0'
<lbl12.example.com> EXEC /bin/sh -c '/usr/bin/python /root/.ansible/tmp/ansible-tmp-1474561274.87-25040398694960/bigip_user.py; rm -rf "/root/.ansible/tmp/ansible-tmp-1474561274.87-25040398694960/" > /dev/null 2>&1 && sleep 0'
An exception occurred during task execution. The full traceback is:
  File "/tmp/ansible_lYiZgE/ansible_module_bigip_user.py", line 714, in <module>
    main()
  File "/tmp/ansible_lYiZgE/ansible_module_bigip_user.py", line 689, in main
    except AdminRoleNoModifyError:
NameError: global name 'AdminRoleNoModifyError' is not defined

fatal: [lbl12.example.com]: FAILED! => {
    "changed": false,
    "failed": true,
    "invocation": {
        "module_name": "bigip_user"
    },
    "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_lYiZgE/ansible_module_bigip_user.py\", line 714, in <module>\n    main()\n  File \"/tmp/ansible_lYiZgE/ansible_module_bigip_user.py\", line 689, in main\n    except AdminRoleNoModifyError:\nNameError: global name 'AdminRoleNoModifyError' is not defined\n",
    "module_stdout": "",
    "msg": "MODULE FAILURE"
}
        to retry, use: --limit @/playbooks/ltm_bootstrap.retry
```

wojtek0806:
Problem:

bigip_user lib needed an overhaul to be usable, this has led to quite a few issues

Analysis:

Reworked bigip_user lib with the corresponding tests and docs, to be in line with the new convention. Added few safety checks to ensure that existing data does not get overwritten.

Tests:

Unit
Functional
Flake 8

### DIFF
--- a/library/bigip_user.py
+++ b/library/bigip_user.py
@@ -360,6 +360,10 @@ class ModuleManager(object):
         Raise when shell attribute is set to 'bash' with roles set to
         either 'admin' or 'resource-admin'.
 
+        NOTE: Admin and Resource-Admin roles automatically enable access to
+        all partitions, removing any other roles that the user might have
+        had. There are few other roles which do that but those roles,
+        do not allow bash.
         """
         if getattr(self.want, 'shell') == 'none' \
                 and getattr(self.have, 'shell') is None:
@@ -381,11 +385,11 @@ class ModuleManager(object):
                 access_have = []
             if access_want:
                 for access in access_want:
-                    if access not in permit:
+                    if access['role'] not in permit:
                         raise F5ModuleError(err)
             if access_have:
                 for access in access_have:
-                    if access not in permit:
+                    if access['role'] not in permit:
                         raise F5ModuleError(err)
 
     def create(self):
--- a/library/bigip_user.py
+++ b/library/bigip_user.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 #
-# Copyright 2016 F5 Networks Inc.
+# Copyright 2017 F5 Networks Inc.
 #
 # This file is part of Ansible
 #
@@ -32,14 +32,6 @@ description:
   - Manage user accounts and user attributes on a BIG-IP.
 version_added: "2.2"
 options:
-  append:
-    description:
-      - If C(yes), will only add groups, not set them to just the list
-        in groups.
-    choices:
-      - yes
-      - no
-    default: no
   full_name:
     description:
       - Full name of the user.
@@ -49,10 +41,10 @@ options:
       - Name of the user to create, remove or modify.
     required: true
     aliases:
-      - user
+      - name
   password_credential:
     description:
-      - Optionally set the users password to this unencrypted value.
+      - Set the users password to this unencrypted value.
         C(password_credential) is required when creating a new account.
     default: None
     required: false
@@ -68,6 +60,7 @@ options:
   partition_access:
     description:
       - Specifies the administrative partition to which the user has access.
+        C(partition_access) is required when creating a new account.
         Should be in the form "partition:role". Valid roles include
         C(acceleration-policy-editor), C(admin), C(application-editor), C(auditor)
         C(certificate-manager), C(guest), C(irule-manager), C(manager), C(no-access)
@@ -75,9 +68,8 @@ options:
         and C(web-application-security-editor). Partition portion of tuple should
         be an existing partition or the value 'all'.
     required: false
-    default: "all:no-access"
+    default: None
     type: list
-    choices: []
   state:
     description:
       - Whether the account should exist or not, taking action if the state is
@@ -89,22 +81,23 @@ options:
       - absent
   update_password:
     description:
-      - C(always) will update passwords if they differ. C(on_create) will only
-        set the password for newly created users.
+      - C(always) will allow to update passwords if the user choses to do so.
+        C(on_create) will only set the password for newly created users.
     required: false
-    default: always
+    default: on_create
     choices:
       - always
       - on_create
 notes:
    - Requires the requests Python package on the host. This is as easy as
      pip install requests
-   - Requires BIG-IP versions >= 13.0.0
+   - Requires BIG-IP versions >= 12.0.0
 extends_documentation_fragment: f5
 requirements:
   - f5-sdk
 author:
   - Tim Rupp (@caphrim007)
+  - Wojciech Wypior (@wojtek0806)
 '''
 
 EXAMPLES = '''
@@ -119,7 +112,6 @@ EXAMPLES = '''
       partition_access: "all:admin"
       state: "present"
   delegate_to: localhost
-
 - name: Change the user "johnd's" role and shell
   bigip_user:
       server: "lb.mydomain.com"
@@ -130,7 +122,6 @@ EXAMPLES = '''
       shell: "tmsh"
       state: "present"
   delegate_to: localhost
-
 - name: Make the user 'johnd' an admin and set to advanced shell
   bigip_user:
       server: "lb.mydomain.com"
@@ -141,7 +132,6 @@ EXAMPLES = '''
       shell: "bash"
       state: "present"
   delegate_to: localhost
-
 - name: Remove the user 'johnd'
   bigip_user:
       server: "lb.mydomain.com"
@@ -150,7 +140,6 @@ EXAMPLES = '''
       name: "johnd"
       state: "absent"
   delegate_to: localhost
-
 - name: Update password
   bigip_user:
       server: "lb.mydomain.com"
@@ -159,6 +148,7 @@ EXAMPLES = '''
       state: "present"
       username_credential: "johnd"
       password_credential: "newsupersecretpassword"
+      update_password: "always"
   delegate_to: localhost
 '''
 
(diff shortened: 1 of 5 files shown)


wojtek0806:
@caphrim007 ready for you

wojtek0806:
found a bug, fix to follow shortly

library/bigip_user.py
```diff
@@ -182,411 +172,368 @@
...
+
+        Raise when shell attribute is set to 'bash' with roles set to
+        either 'admin' or 'resource-admin'.
+
+        NOTE: Admin and Resource-Admin roles automatically enable access to
+        all partitions, removing any other roles that the user might have
+        had. There are few other roles which do that but those roles,
+        do not allow bash.
+        """
+        if getattr(self.want, 'shell') == 'none' \
```
caphrim007: dont use getattr. these should be self.want.shell

library/bigip_user.py
```diff
@@ -182,411 +172,368 @@
...
+                and getattr(self.have, 'shell') is None:
```
caphrim007: self.have.shell

library/bigip_user.py
```diff
@@ -182,411 +172,368 @@
...
+            error = "Attribute 'bash' is already set to 'none' on target " \
+                    "device."
+            raise F5ModuleError(error)
+
+        if getattr(self.want, 'shell') == 'bash':
```
caphrim007: self.want.shell

library/bigip_user.py
```diff
@@ -182,411 +172,368 @@
...
+            err = "Shell access is only available to 'admin' or " \
+                  "'resource-admin' roles"
+            permit = ['admin', 'resource-admin']
+            if self.want is not None:
```
caphrim007: want will never be none. it's supplied from the user

library/bigip_user.py
```diff
@@ -182,411 +172,368 @@
...
+                access_want = getattr(self.want, 'partition_access')
```
caphrim007: self.want.partition_access

library/bigip_user.py
```diff
@@ -182,411 +172,368 @@
...
-            result['description'] = self.params['full_name']
-        if self.is_password_changed():
-            result['password'] = self.params['password_credential']
-        if self.is_shell_changed(current):
-            result['shell'] = self.params['shell']
-        if self.is_partition_access_changed(current):
-            result['partitionAccess'] = self.getChangedPartitionAccess()
-        return result
+                access_want = []
+            if self.have is not None:
```
caphrim007: have will never be none. it's taken from the rest api.

library/bigip_user.py
```diff
@@ -182,411 +172,368 @@
...
+                access_have = getattr(self.have, 'partition_access')
```
caphrim007: self.have.partition_access

library/bigip_user.py
wojtek0806: it will be None, as this method is used at create too

roles/bigip_user/tasks/test-post-version-12-specific.yaml
```diff
@@ -17,23 +19,45 @@
...
+  register: result
+
+- name: Assert Update password for user on a different partition
+  assert:
+      that:
+          - result|changed
+
 - name: Remove user on different partition
   bigip_user:
       server: "{{ inventory_hostname }}"
```
caphrim007: somewhere in the functional test file, add the following tests

* users on different partitions. there was a bug in pre-13 that didnt allow exists() to work for this example
* users specifying multiple different partition_access for different partitions
* users specifying multiple different partition_access for same partitions

for example

partition: bar
partition_access:
    - Common:operator
    - bar:user_manager


partition_access:
    - Common:operator
    - Common:user_manager


partition_access:
    - Common:operator
    - bar:resource_admin


i expect some of the above to fail, so make sure the failures coming from bigip are correct.

library/bigip_user.py
```diff
@@ -182,411 +172,366 @@
...
-            return self.load_user_without_partition()
-        else:
-            return self.load_user_with_partition()
+    def api_params(self):
+        result = {}
+        for api_attribute in self.api_attributes:
+            if api_attribute in self.api_map:
+                result[api_attribute] = getattr(
+                    self, self.api_map[api_attribute])
+            elif api_attribute == 'password':
```
wojtek0806: you cannot do that, i have already tried that, this causes bigip password to be written here if i remove this

library/bigip_user.py
```diff
@@ -182,411 +172,366 @@
...
+                result[api_attribute] = self._values['password_credential']
+            elif api_attribute == 'name':
```
wojtek0806: i think i placed it here as i was getting problems with 'name' being an attribute of self.want, this was when i had different variation of the current code, so i can try to see if this cases tests to fail

library/bigip_user.py
```diff
@@ -182,411 +172,366 @@
...
+            raise F5ModuleError(err)
+        if self.want.partition_access is None:
+            err = "The 'partition_access' option " \
+                  "is required when creating a resource."
+            raise F5ModuleError(err)
+
+    def validate_shell_parameter(self):
+        """Method to validate shell parameters.
+
+        Raise when 'none' is being set for shell parameter,
```
wojtek0806: why not? so how do you inform the user what was wrong? this behavior is specific to bigip user. shell can be 'none' but we won't get that from device. unless you expect the users to know the intricacies of bigip operation

library/bigip_user.py
```diff
@@ -182,411 +172,366 @@
...
+            if self.have is not None:
+                access_have = self.have.partition_access
             else:
-                r = api.tm.auth.users.user.load(
-                    name=self.params['username_credential'],
-                    partition=self.params['partition']
-                )
-                r.modify(**params)
+                access_have = []
+            if access_want:
```
wojtek0806: well, that is not true, as any for partition name does not equal to admin, resource-admin. It can be few other roles which do not allow bash

library/bigip_user.py
caphrim007: because noops should not raise errors. it's a foundation of devops. you only need to inform the user of what was wrong if they did something wrong. specifying a noop is not doing anything wrong.

you should be setting it to 'none' for them when you receive nothing from the device.

caphrim007: write a test to set the shell, then set it to none and then idempotent set it to none.  it should not fail, otherwise it is not idempotent

roles/bigip_user/tasks/main.yaml
```diff
@@ -159,25 +198,27 @@
...
   assert:
       that:
-          - not result|changed
+          - result|changed
 
-# This should fail because the user does not have the appropriate role
-- name: Update user shell to bash
+# This will fail as BIGIP does not return 'shell' attribute if it is already
+# set to 'none'
+- name: Update user shell to none - Idempotent check
```
caphrim007: this should not fail. fix the code

wojtek0806: it was failing due to exception being raised :)

library/bigip_user.py
wojtek0806: yep, when i remove this i get the following error:

````
fatal: [ltm3]: FAILED! => {"changed": false, "failed": true, "module_stderr": "/home/drgun/virtenvs/ansible/local/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py:852: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings\n  InsecureRequestWarning)\n/home/drgun/virtenvs/ansible/local/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py:852: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings\n  InsecureRequestWarning)\n/home/drgun/virtenvs/ansible/local/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py:852: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings\n  InsecureRequestWarning)\nTraceback (most recent call last):\n  File \"/home/drgun/temp/ansible_ilseij/ansible_module_bigip_user.py\", line 536, in <module>\n    main()\n  File \"/home/drgun/temp/ansible_ilseij/ansible_module_bigip_user.py\", line 530, in main\n    results = mm.exec_module()\n  File \"/home/drgun/temp/ansible_ilseij/ansible_module_bigip_user.py\", line 310, in exec_module\n    changed = self.present()\n  File \"/home/drgun/temp/ansible_ilseij/ansible_module_bigip_user.py\", line 325, in present\n    return self.create()\n  File \"/home/drgun/temp/ansible_ilseij/ansible_module_bigip_user.py\", line 398, in create\n    self.create_on_device()\n  File \"/home/drgun/temp/ansible_ilseij/ansible_module_bigip_user.py\", line 432, in create_on_device\n    **params)\nTypeError: create() got multiple values for keyword argument 'name'\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 0}

````

File: library/bigip_user.py (hunk 9 above)
Comment (wojtek0806) [inline library/bigip_user.py]: here is q output:
````
 1.3s __init__: self.want.password='admin'

 1.3s __init__: self.want.password='admin'
 1.4s api_params: self._values['password_credential']='password'

 1.3s __init__: self.want.password='admin'
````

Comment (wojtek0806) [inline library/bigip_user.py]: test user pass is : password
api user pass is: admin

File: library/bigip_user.py (hunk 10 above)
Comment (wojtek0806) [inline library/bigip_user.py]: this causes the **params that is passed to:
```
            self.client.api.tm.auth.users.user.create(name=self.want.name,
                                                      **params)
```
to already contain name=, i can remove name=self.want.name, but I wanted to stay consistent with the way we have this bit in other modules

File: library/bigip_user.py (hunk 11 above)
Comment (wojtek0806) [inline library/bigip_user.py]: ok, so i understand you are ok with us providing a default? or is this only in such cases where we need to ensure that idempotent test pass ?

File: library/bigip_user.py (hunk 12 above)
Comment (wojtek0806) [inline library/bigip_user.py]: now i see what you mean :) ok, please use gist easier to read

File: library/bigip_user.py [hunk 14]
```diff
@@ -182,411 +172,366 @@
...
+                for access in access_want:
+                    if access['role'] not in permit:
+                        raise F5ModuleError(err)
+            if access_have:
```
wojtek0806: ok

library/bigip_user.py
```diff
@@ -182,411 +172,366 @@
...
+            if self.want.shell == 'none' and self.have.shell is None:
+                error = "Attribute 'bash' is already set to 'none' " \
+                        "on target device."
+                raise F5ModuleError(error)
+
+        if self.want.shell == 'bash':
+            err = "Shell access is only available to 'admin' or " \
+                  "'resource-admin' roles"
+            permit = ['admin', 'resource-admin']
+            access_want = self.want.partition_access
```
wojtek0806: ok

library/bigip_user.py
```diff
@@ -182,411 +172,366 @@
...
+            # This can occur when we do checks at create
+            if self.have is not None:
```
wojtek0806: we will error out if self.have is a None object

library/bigip_user.py
wojtek0806: what are we checking here? we are checking if we have 'role' in partition_access, where are we checking if the role is either admin, or resource-admin ?

library/bigip_user.py
caphrim007: no default needs to be provided. if the rest api, upon create, gets no shell, then it set it to whatever it wants. If, upon update, you get no shell attribute from the api, then you should assume that means the "none" shell.

It's not much different that what I do for static routes here

https://github.com/F5Networks/f5-ansible/blob/devel/library/bigip_static_route.py#L231

bigip is giving me the word "default" so i transform that to the IP address because I am expecting IP addresses from the user

library/bigip_user.py
caphrim007: it was an example. just collapse it into an any expression

`x['role'] if x['role'] in permit for ....`

or whatever it becomes

roles/bigip_user/tasks/main.yaml
```diff
@@ -1,22 +1,44 @@
...
-      update_password: "on_create"
   register: result
 
 - name: Assert Create user
   assert:
       that:
           - result|changed
 
+
 - name: Create user - Idempotent check
```
caphrim007: the idempotent check needs to reflect, exactly, the options specified in the test that it is checking for itempotency. so this test needs to be identical to the one above it.

roles/bigip_user/tasks/main.yaml
```diff
@@ -225,5 +265,5 @@
   register: result
 
 - name: Include version 13 (Daytona) checks
-  include: test-version-13-specific.yaml
+  include: test-post-version-12-specific.yaml
   when: system_info.product_information.product_version >= "13.0.0"
```
caphrim007: somewhere in this functional test file can we test for two things, unless I missed them,

1. what happens if a new user is created without specifying a partition_access
2. how does an admin create a user, deliberately, with no access. For example, there is a role called no-access in the api.

library/bigip_user.py
```diff
@@ -182,411 +172,358 @@
...
+        self.remove_from_device()
+        if self.exists():
+            raise F5ModuleError("Failed to delete the user")
+        return True
 
-    def create_user_on_device(self, params):
-        tx = self.api.tm.transactions.transaction
-        with TransactionContextManager(tx) as api:
-            api.tm.auth.users.user.create(**params)
+    def remove_from_device(self):
```
caphrim007: These _on|from_device codes should reflect the pattern that is used in GTM wide_ip.

In that module, you'll find a ModuleManager that is relatively small and which uses a factory method to create different concrete Managers based on the version of BIG-IP it detects.

This allows for two things at least

1. that we can expand this module if they break something in the future without having to worry about breaking the existing versions
2. so that we can customize future, or existing, versions with other executable code if we find that something needs to be refactored or changed.

This is an example of how the module manager would look

https://github.com/F5Networks/f5-ansible/blob/devel/library/bigip_gtm_wide_ip.py#L217

and these are examples of the two specialized classes.

https://github.com/F5Networks/f5-ansible/blob/devel/library/bigip_gtm_wide_ip.py#L336
https://github.com/F5Networks/f5-ansible/blob/devel/library/bigip_gtm_wide_ip.py#L383

Note that they inherit from a BaseManager class that includes code general to each specialized manager.

roles/bigip_user/tasks/main.yaml
wojtek0806: you have to provide something in partition_access, in this case you can provide something like:
none:no-access. But this is mandatory otherwise API will complain, this is true for v12+, in v11, bigip filled in the blanks

roles/bigip_user/tasks/main.yaml
wojtek0806: if i specify password in the create user, then it will fail, as we require update_password to be set to Always, at the moment default is on_create

library/bigip_user.py
wojtek0806: not sure what you are aiming at here? so do you want me to create subclass of ModuleManager ?

wojtek0806: and just put all custom items into that class ?

roles/bigip_user/tasks/main.yaml
wojtek0806: unless we go with 'always' as default and 'on_create' for creation, which is something i was considering

PR #108, opened 2017-03-28
Update installing_modules.rst

pcloup:
On MacOSX, it appears that the directory where ANSIBLE is installing modules is different, as the /usr/local/lib is not accessible for a standard user, even with "sudo".
Not sure if it makes sense to update the doc for this reason, but it can help ;-)


caphrim007:
lgtm

PR #115, opened 2017-04-02
Fixes #90

### LINKED ISSUE #90: refactor bigip_vlan to use new conventions
this is a 2nd gen module that needs refactoring to use 3rd gen conventions

* Include unit tests

wojtek0806:
Problem:

bigip_vlan lib needed an overhaul to use 3rd gen conventions.

Analysis:

Reworked bigip_vlan lib with the corresponding tests and docs, to be in
line with the new convention.

Tests:

Unit
Functional
Flake 8


library/bigip_vlan.py
```diff
@@ -154,300 +152,301 @@
...
+        self.update_on_device()
+        return True
 
+    def remove(self):
+        if self.client.check_mode:
+            return True
+        self.remove_from_device()
         if self.exists():
-            raise F5ModuleError("Failed to delete the VLAN")
+            raise F5ModuleError("Failed to delete the user")
```
caphrim007: "vlan"

library/bigip_vlan.py
```diff
@@ -154,300 +152,301 @@
...
+            return True
+        return False
 
-        if description is not None:
-            if 'description' in current:
-                if description != current['description']:
-                    params['description'] = description
-            else:
-                params['description'] = description
+    def _parse_interfaces(self, key):
```
caphrim007: this method is too long. it needs to be broken up into smaller testable pieces

library/bigip_vlan.py
```diff
@@ -154,300 +152,301 @@
...
+        """Operate on interfaces during setup/update.
+
+
+        This method can be executed ruing create or update()
```
caphrim007: "ruing" ?

library/bigip_vlan.py
```diff
@@ -154,300 +152,301 @@
...
-                        params['interfaces'] = tmp
+                    changed[key] = getattr(self.want, key)
+        if changed:
+            self.changes = Parameters(changed)
+
+    def _update_changed_options(self):
+        changed = {}
+        for key in Parameters.updatables:
+            if getattr(self.want, key) is not None:
+                if key == 'tagged_interfaces' or key == 'untagged_interfaces':
```
caphrim007: this method is supposed to be a baseclass method. do not get in the habit of putting module specific tests here. if you're doing that, then those should go in the parameters class

wojtek0806:
The 'heavy lifting' here needs to have access to items returned by querying device, i have no idea how to stash those objects so it is available outside Module Manager instance, you are operating on sub-collections here, also 'interfaces' parameter is something that is populated after we process either:
'tagged_interfaces' or 'untagged_interfaces', which is required for this to work seamlessly with the old functional tests.

wojtek0806:
i guess i could try to leverage _values, will give it a shot

library/bigip_vlan.py
```diff
@@ -154,300 +152,301 @@
...
-            partition=partition
+    def create(self):
+        self._set_changed_options()
+        if self.client.check_mode:
+            return True
+        self.create_on_device()
+        return True
+
+
+class VlanManager(BaseManager):
```
wojtek0806: well you wanted to subclass, i have not seen a reason for it either, just going by what we have done last time

wojtek0806:
Moved the checking into Parameters, I had to modify _set_changed_options() and _update_changed_options() to pass the result of self.get_interfaces_from_device() as appropriate, so the magic can happen inside parameters.

Much happier with the result now to be honest :)

@caphrim007

caphrim007:
@wojtek0806 you are free to put the client in the parameters class. i have to do that often in the iworfklow modules. for example.

https://github.com/F5Networks/f5-ansible/blob/devel/library/iworkflow_license_pool_member.py#L195

wojtek0806:
trying to understand this bit before i go and use it
 ````
self.client = client
self.want.client = self.client
self.want = Parameters()
self.want.update(self.client.module.params)
 ````
so does this mean that we create another 'client' instance inside Parameters ?
````
self.want.client = self.client
`````

wojtek0806:
opening new PR

PR #206, opened 2017-05-19
Update bigip_virtual_server2.py

aabond:
When trying to ensure a vServer is present, the self function is trying to validate existence of virtual server with the arguments self.have which fails.  In tracing the error, it seems the self.want would be the appropriate arguments to pass to check for the existence of the desired vServer.  I've tested this myself and it appears to be working now.


caphrim007:
lgtm

PR #225, opened 2017-05-30
Get the device, not the device group

nahun:
The function was retrieving the device group. Probably just leftover from using the device group module as a template is my guess.


caphrim007:
lgtm

PR #226, opened 2017-05-30
force-full-load-push and Not All Devices Synced state

nahun:
The overwrite_config parameter should add the force-full-load-push argument to the sync command unless I'm misunderstanding its purpose.

Need to wait for "Not All Devices Synced" state as well since the device group will typically go into this state immediately after starting the sync.

### DIFF
--- a/library/bigip_configsync_actions.py
+++ b/library/bigip_configsync_actions.py
@@ -149,6 +149,13 @@ class Parameters(AnsibleF5Parameters):
         return result
 
     @property
+    def force_full_push(self):
+        if self.overwrite_config:
+            return 'force-full-load-push'
+        else:
+            return ''
+
+    @property
     def overwrite_config(self):
         result = self._cast_to_bool(self._values['overwrite_config'])
         return result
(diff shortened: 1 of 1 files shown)


caphrim007:
i am certainly not the expert in HA pair stuff, so I would seem to think that yes, that's how its intended to behave. I think that option was there to mimic a checkbox in the UI

nahun:
The docs don't say for sure its the same thing as the checkbox in the UI, but I'm pretty sure it is as well.

Also, be aware the force-full-load-push option was added in 12.1 so it wouldn't work in anything previous to that: [https://support.f5.com/csp/article/K63470472](url)

caphrim007:
ok, then we'll need to have a different form of module manager. thanks for pointing that out. i'll merge this and then refactor the module manager.

caphrim007:
do you know if it fails with an "unknown option" or something if that option is specified on a version less than 12.1?

nahun:
tmsh just shows:
```
(tmos)# run cm config-sync to-group CIS_LTM_LOCAL_SEA force-full-load-push
Syntax Error: unexpected argument "force-full-load-push"
```
I haven't tried the REST API

caphrim007:
the rest api would do the same because we're essentially calling `tmsh` over rest. merging so i can refactor

PR #249, opened 2017-06-19
Bigip software new concept

wojtek0806:
New. Concept  module


wojtek0806:
@tar

wojtek0806:
@caphrim007 this is now ready for review

library/bigip_software.py
```diff
@@ -18,12 +18,6 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
-ANSIBLE_METADATA = {
```
caphrim007: this is a required variable. please re-add it

library/bigip_software.py
```diff
@@ -46,64 +40,148 @@
...
+      - The volume to install the software and, optionally, the hotfix to. This
+        parameter is only required when the C(state) is either C(activated) or
+        C(installed).
+    required: false
   software:
     description:
       - The path to the software (base image) to install. The parameter must be
         provided if the C(state) is either C(installed) or C(activated).
+      - If C(remote_src) is used, this parameter will have to be an 
+        C(unauthenticated) C(HTTP) or C(HTTPS)link to the software image.
```
caphrim007: does it need to be unauthenticated? what happens if I use a url with basic auth credentials in it?

ex.
https://admin:pass@site.local.com/path/to/image.iso

library/bigip_software.py
```diff
@@ -46,64 +40,148 @@
...
+      - C(Caveat): When providing link to the software ISO, 
+        if the ISO name is different than the one listed inside the 
+        C(software_md5) md5sum file. We will change it accordingly when 
+        saving the files on the device. This might lead to ISO names not 
+        matching the links provided in C(software).
```
caphrim007: make sure you provide an example of this in the EXAMPLES variable to illustrate it.

library/bigip_software.py
```diff
@@ -46,64 +40,148 @@
...
-        module will create HD1.2 and install the software. If volume name does not end with
-        numeric character, then add .1 to the current active volume name. When C(volume)
-        is specified, this option will be ignored.
+      - Automatically chooses the first inactive volume in alphanumeric order. 
+        If there is no inactive volume, new volume with incremented volume name 
+        will be created. For example, if HD1.1 is currently active and no other 
+        volume exists, then the module will create HD1.2 and install the 
+        software. If volume name does not end with numeric character, 
+        then add .1 to the current active volume name. When C(volume) is 
+        specified, this option will be ignored.
```
caphrim007: make sure you provide an example for this in the EXAMPLES variable to illustrate this

library/bigip_software.py
```diff
@@ -46,64 +40,148 @@
...
+  hotfix:
     description:
-      - The volume to install the software and, optionally, the hotfix to. This
-        parameter is only required when the C(state) is either C(activated) or
-        C(installed).
+      - The path to an optional Hotfix to install. This parameter requires that
+        the C(software) parameter be specified or the corresponding software 
+        image exists on the unit.
+      - If C(remote_src) is used, this parameter will have to be an 
+        unauthenticated http or https link to the hotfix image.
```
caphrim007: should "unauthenticated", "http" and "https" use the same syntax C(*) as above?

library/bigip_software.py
```diff
@@ -46,64 +40,148 @@
...
+      - hotfix_image
+  software_md5sum:
+    description:
+      - The link to an MD5 sum file of the remote software ISO image, 
+        it is required when  C(software) parameter is used and C(remote_src) 
+        is selected. 
+      - Parameter only used when and C(state) is: C(installed), C(activated) 
+        or C(present).
+    required: false
+    default: None
```
caphrim007: you can remove all lines that are

"required: false"
"default: None"

this is a new request from upstream vendor, so I've begun making the switch

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-import os
-import time
+ANSIBLE_METADATA = {
+    'status': ['preview'],
+    'supported_by': 'community',
+    'metadata_version': '1.0'
+}
+
 import io
+from ansible.module_utils.f5_utils import *
```
caphrim007: ansible now requires that this not be an import *.

expand it to import the classes you need.

from ansible.module_utils.f5_utils import (
    foo1,
    foo2
)

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+      password: "admin"
+      hotfix: "http://fake.com/Hotfix-12.1.2.1.0.271-HF1.iso"
+      hotfix_md5sum: "http://fake.com/Hotfix-12.1.2.1.0.271-HF1.iso.md5"
+      software: "http://fake.com/BIGIP-12.1.2.0.0.249.iso"
+      software_md5sum: "http://fake.com/BIGIP-12.1.2.0.0.249.iso.md5"
+      build: "1.0.271"
+      version: "12.1.2"
+      remote_src: "yes"
+      state: "installed"
+      volume: "HD1.2"
```
caphrim007: provide more examples than just remote examples

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+      server: "bigip.localhost.localdomain"
+      user: "admin"
+      password: "admin"
+      software: "http://fake.com/BIGIP-12.1.2.0.0.249.iso"
+      software_md5sum: "http://fake.com/BIGIP-12.1.2.0.0.249.iso.md5"
+      build: "0.0.249"
+      version: "12.1.2"
+      remote_src: "yes"
+      volume: "HD1.1"
+      state: "installed"
```
caphrim007: requires delegate_to: localhost for all examples

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+      hotfix: "http://fake.com/Hotfix-12.1.2.1.0.271-HF1.iso"
+      hotfix_md5sum: "http://fake.com/Hotfix-12.1.2.1.0.271-HF1.iso.md5"
+      software: "http://fake.com/BIGIP-12.1.2.0.0.249.iso"
+      software_md5sum: "http://fake.com/BIGIP-12.1.2.0.0.249.iso.md5"
+      build: "1.0.271"
+      version: "12.1.2"
+      remote_src: "yes"
+      state: "installed"
+      volume: "HD1.2"
 '''
```
caphrim007: You need a RETURN variable that describes all of the returnables you specify

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+    @property
+    def remote_src(self):
+        remote = self._values['remote_src']
+        image_md5sum = self._values['software_md5sum']
+        hotfix_md5sum = self._values['hotfix_md5sum']
+        state = self._values['state']
+        hotfix = self._values['hotfix']
+        software = self._values['software']
+        err = 'You must provide md5sum file links ' \
+              'if using remote source.'
```
caphrim007: this can be one line. ansible's column limit is 120

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-            version=None,
-            build=None
-        )
+    @property
+    def volume(self):
+        state = self._values['state']
+        volume = self._values['volume']
+        reuse = self._values['reuse_inactive_volume']
+
+        if state == 'activated' or state == 'installed':
```
caphrim007: avoid using a large if block like this. instead do

if state in ['absent', 'present']:
    return self._values['volume']

and then backspace everything that was formerly under the if statement.

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+            if volume is not None:
+                if not self._volume_exists(volume):
+                    self._values['create_volume'] = True
+                    return self._values['volume']
+                else:
+                    self._check_active_volume(volume)
+                    return self._values['volume']
+
+            if volume is None:
+                if state == 'activated' or state == 'installed':
```
caphrim007: previous comment applies here too

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+                        if self._volume_exists(volume):
+                            self._values['delete_volume'] = True
+                            self._values['create_volume'] = True
+                        else:
+                            self._values['create_volume'] = True
+                        self._values['volume'] = volume
+                        return self._values['volume']
+                    else:
+                        err = "You must specify a volume."
+                        raise F5ModuleError(err)
```
caphrim007: I prefer error messages just part of the exception class. dont bother using a variable

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+    def _find_iso_content(self, iso):
+        paths = ['/METADATA.XML', 'metadata.xml']
 
+        for path in paths:
+            try:
+                content = iso.record(path).content
+                return content
+            except KeyError:
+                pass
+        raise Exception(
```
caphrim007: F5ModuleError should be used here

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        name = source.path.split('/')[-1]
+        if source.scheme:
+            if source.scheme == 'http':
+                return name
+            elif source.scheme == 'https':
+                self._values['secure'] = True
+                return name
+            else:
+                err = 'Only remote HTTP or HTTPS sources are supported.'
+                raise F5ModuleError(err)
```
caphrim007: avoid the variable. just put the message in the exception instantiation

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-            software=[],
-            hotfix=[]
-        )
+    def _load_volume(self, volume):
+        vol = self.client.api.tm.sys.software.volumes.volume.load(
+            name=volume)
+        return vol
+
+    def _volume_exists(self, volume):
+        return self.client.api.tm.sys.software.volumes.volume.exists(
```
caphrim007: assign a variable to the return value before returning it. this helps debugging later

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+                return name
+            else:
+                err = 'Only remote HTTP or HTTPS sources are supported.'
+                raise F5ModuleError(err)
 
-        self.api = bigip_api(kwargs['server'],
-                             kwargs['user'],
-                             kwargs['password'],
-                             kwargs['validate_certs'])
+    def _list_volumes(self):
```
caphrim007: since these are methods whose sole purpose is to communicate with the device, the pattern we use for that is naming them *_on_device or *_from_device. I would change these to match that pattern

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        volumes = self.client.api.tm.sys.software.volumes.get_collection()
+        return volumes
 
-    def read(self):
-        result = dict(
-            software=[],
-            hotfix=[]
-        )
+    def _load_volume(self, volume):
```
caphrim007: match *_on_device or *_from_device pattern

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        vol = self.client.api.tm.sys.software.volumes.volume.load(
+            name=volume)
+        return vol
+
+    def _volume_exists(self, volume):
```
caphrim007: match *_on_device or *_from_device pattern

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        return self.client.api.tm.sys.software.volumes.volume.exists(
+            name=volume)
+
+    def _check_active_volume(self, volume):
+        v = self._load_volume(volume)
+        if hasattr(v, 'active') and v.active is True:
+            err = "Cannot install software or hotfixes to active volumes."
+            raise F5ModuleError(err)
```
caphrim007: error message should go straight into the exception

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+
+    def _get_next_available_volume(self):
+        volumes = self._list_volumes()
+        if len(volumes) == 1:
```
caphrim007: this if statement is redundant as it and its else do the exact same thing; loop over volumes.

just combine the for loops and eliminate the if here

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        self.want.update(self.client.module.params)
 
-        if target_volume is None:
-            try:
-                _active_volume = active_volume.split('.')
-                _active_volume[-1] = str(int(_active_volume[-1]) + 1)
-                target_volume = '.'.join(_active_volume)
-            except Exception:
-                target_volume = active_volume + '.1'
+    def exec_module(self):
```
caphrim007: i think your logic is incorrect here. the logic implies that I can only use remote sources or only use local sources. however, what happens if I use one remote source and one local source. Additionally where is this enforced?

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        else:
+            self.install_volume()
+            self.wait_for_device_reboot()
+            return True
 
-        with open(filename, 'rb') as fh:
-            remote_path = "/shared/images/%s" % os.path.basename(filename)
+    def is_activated(self):
+        result = self.software_on_volume()
+        if result:
```
caphrim007: i would test the negation here, return none if it succeeds, and then backspace the code below it.

if not result:
    return False

self.have.....

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-                        chain_type = 'FILE_MIDDLE'
-
-                self.api.System.ConfigSync.upload_file(
-                    file_name=remote_path,
-                    file_context=dict(
-                        file_data=text,
-                        chain_type=chain_type
-                    )
-                )
+    def is_installed(self):
```
caphrim007: see comment for is_activated for this as well.

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        if not self.exists():
+            self.upload()
+        if self.want.delete_volume:
+            self.delete_volume_on_device()
+        if self.want.hotfix:
+            version = self.want.version
+            if self.base_image_exists():
+                self.install_hotfix_on_device()
+            else:
+                err = 'Base image of version: {0} must exist to install ' \
```
caphrim007: exception can go in the object instantiation

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-        )
+    def install_volume(self):
+        self._set_changed_options()
+        if self.client.check_mode:
+            return True
+        if not self.exists():
+            self.upload()
+        if self.want.delete_volume:
+            self.delete_volume_on_device()
+        if self.want.hotfix:
```
caphrim007: test the negation  you do on line 825 and return early. then backspace all of the code that follows here

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-                    pver = psoftware['version']
-                    pprod = psoftware['product']
+        if forced:
+            return False
 
-                    if pbuild == ibuild and pver == iver and pprod == iprod:
-                        result = True
+        if software_path and hotfix_path:
+            if self.image_exists_on_device() and \
+             self.hotfix_exists_on_device():
```
caphrim007: this can be on the previous line. 120 columns

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+                return True
+        elif hotfix_path:
+            if self.hotfix_exists_on_device():
+                return True
+        elif software_path:
+            if self.image_exists_on_device():
+                return True
```
caphrim007: there is no false to catch if this fails. The else at the end of the if block is not sufficient.

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        else:
```
caphrim007: eliminate this else and return false for anything that lands at the end of this method

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-            )
+    def list_volumes_on_device(self):
+        volumes = self.client.api.tm.sys.software.volumes.get_collection()
+        return volumes
 
-        images = self.read()
-        total = sum(len(v) for v in images.itervalues())
+    def image_exists_on_device(self):
+        images = self.client.api.tm.sys.software.images.get_collection()
+        for image in images:
```
caphrim007: Since this code is repeated over and over again, i would extract it out and change this for loop into
```
if any(self._has_build_and_version(i) for i in images):
    return True
```
then define the _has_build method above as
```
if image.version != self.want.version
    return False
if image.build != self.want.build
    return False
return True
```

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        return False
 
-        if force:
-            if hotfix:
-                self.delete(hotfix)
-                total -= 1
-                changed = True
+    def hotfix_exists_on_device(self):
+        hotfixes = self.client.api.tm.sys.software.hotfix_s.get_collection()
+        for image in hotfixes:
```
caphrim007: see comment above

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+            if version is None or build is None:
+                if hotfix:
+                    self.set_product_info(True)
+                else:
+                    self.set_product_info()
+            return False
 
-        status = self.api.System.SoftwareManagement.get_all_software_status()
-        volumes = [x['installation_id']['install_volume'] for x in status if x['active']]
+        if state == 'present' or state == 'absent':
```
caphrim007: this state and the above are mutually exclusive, so make this an `elif`

also change the statement to be if state in ['presen...']

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+                    tmp_res = volume.attrs
+                    return Parameters(tmp_res), volume
+
+    def check_product_info(self):
+        hotfix = self.want.hotfix
+        version = self.want.version
+        build = self.want.build
+        state = self.want.state
+
+        if state == 'installed' or state == 'activated':
```
caphrim007: change statement to be `if state in ['instal....']`

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+            info = self.load_hotfix_from_device_by_name()
         else:
-            create_volume = True
-
+            info = self.load_image_from_device_by_name()
+
+        if info is None:
+            err = 'Unable to set ISO information based on ISO name, ' \
+                  'please specify version and build.'
+            raise F5ModuleError(err)
```
caphrim007: message in module error. not variable

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-            self.wait_for_images(total)
+            md5name = self.want.hfmd5_name
+            md5_url = self.want.hotfix_md5sum
+            name = self.name_match(md5_url, md5name, True)
+            url = self.want.hotfix
+            cmd = self.prepare_command(url, name)
+            self.run_command_on_device(cmd)
+            if not self.does_checksum_match(md5name, name):
+                err = 'ISO download failed from remote host.'
+                raise F5ModuleError(err)
```
caphrim007: error in object, not variable

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        else:
+            md5name = self.want.imgmd5_name
+            md5_url = self.want.software_md5sum
+            name = self.name_match(md5_url, md5name)
+            url = self.want.software
+            cmd = self.prepare_command(url, name)
+            self.run_command_on_device(cmd)
+            if not self.does_checksum_match(md5name, name):
+                err = 'ISO download failed from remote host.'
+                raise F5ModuleError(err)
```
caphrim007: error in object, not variable

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        if hotfix:
+            if name != self.want.hotfix_name:
+                self.want.hotfix_name = name
+        else:
+            if name != self.want.software_name:
+                self.want.software_name = name
 
-        status = self.api.System.SoftwareManagement.get_all_software_status()
-        volumes = [x['installation_id']['install_volume'] for x in status]
+    def load_image_from_device_by_name(self):
```
caphrim007: to stick with the *_on_device pattern, this should be renamed to load_image_by_name_from_device

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        if exists:
+            result = self.client.api.tm.sys.software.images.image.load(
+                name=self.want.software_name
+            )
+        return result
 
-        if hotfix:
-            photfix = self.params['photfix']
-            self.install_software(photfix, reboot=False, create=False)
+    def load_hotfix_from_device_by_name(self):
```
caphrim007: load_hotfix_by_name_from_device

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+                self.want.hotfix_name = name
+        else:
+            if name != self.want.software_name:
+                self.want.software_name = name
 
-        status = self.api.System.SoftwareManagement.get_all_software_status()
-        volumes = [x['installation_id']['install_volume'] for x in status]
+    def load_image_from_device_by_name(self):
+        result = None
+        exists = self.image_exists_on_device_by_name()
```
caphrim007: image_exists_by_name_on_device

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
-    def absent(self):
-        hotfix = self.params['hotfix']
-        software = self.params['software']
+    def run_command_on_device(self, cmd):
+        err = 'Could not execute command. Most likely device is unresponsive.'
+        result = self.client.api.tm.util.bash.exec_cmd('run', utilCmdArgs=cmd)
+        if result:
+            return result
+        else:
+            raise F5ModuleError(err)
```
caphrim007: use text of error message instead of a variable

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+            if hotfix.version == self.want.version:
+                if hotfix.build == self.want.build:
+                    self.name_check(hotfix.name, True)
+                    return True
+        return False
 
-        if hotfix:
-            self.delete(hotfix)
-            total -= 1
+    def hotfix_exists_on_device_by_name(self):
```
caphrim007: hotfix_exists_by_name_on_device

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+        return self.client.api.tm.sys.software.hotfix_s.hotfix.exists(
+            name=self.want.hotfix_name
+        )
 
-        if software:
-            self.delete(software)
-            total -= 1
+    def image_exists_on_device_by_name(self):
```
caphrim007: image_exists_by_name_on_device

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+    def load_volume_on_device(self):
+        volume = self.client.api.tm.sys.software.volumes.volume.load(
+            name=self.want.volume)
+        return volume
+
+    def reboot_volume_on_device(self):
+        self.client.api.tm.sys.software.volumes.exec_cmd(
+            'reboot', volume=self.want.volume
+            )
+    def volume_exists_on_device(self):
```
caphrim007: missing a newline above this method

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+    def __init__(self):
+        self.states = ['absent', 'activated', 'installed', 'present']
+        self.supports_check_mode = True
+        self.argument_spec = dict(
+            state=dict(
+                default='activated',
+                choices=self.states
+            ),
+            force=dict(
+                required=False,
```
caphrim007: remove

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+                type='bool',
+                default='no'
+            ),
+            hotfix=dict(
+                required=False,
```
caphrim007: remove

library/bigip_software.py
```diff
@@ -212,673 +290,1060 @@
...
+                aliases=['hotfix_image'],
+                default=None
```
caphrim007: remove all `default: None` and `required: False` because those are the defaults of argumentspec

library/bigip_software.py
wojtek0806: this needs to stay as it is now, i would need to add more parameters for authenticated sources like user and password.

library/bigip_software.py
wojtek0806: what do you mean ?

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ?

library/bigip_software.py
wojtek0806: what do you mean? you have both local source and remote source examples

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: they are mutually exclusive, this will not work, that is how it should work, you cannot combine remote source and local source sorry, this adds too much complexity and a whole set of new problems and i don't intend to have it like that

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: k

library/bigip_software.py
wojtek0806: ok

library/bigip_software.py
wojtek0806: choices need to be there, as your state Activated is default, not Present like in other modules so what do you want to remove?

library/bigip_software.py
wojtek0806: ???

library/bigip_software.py
wojtek0806: does this apply to the above comments as it is not clear

library/bigip_software.py
caphrim007: as i mentioned in other comments, required false is the default. this applies to all of the argument spec

library/bigip_software.py
caphrim007: yes

library/bigip_software.py
caphrim007: the comment was for line 1283. github shows a small blob above the comment which i assume you're referring to here

library/bigip_software.py
caphrim007: https://github.com/F5Networks/f5-ansible/blob/devel/library/bigip_pool.py#L185

you're missing the RETURN variable

library/bigip_software.py
caphrim007: i mean none of the examples are using filesystem paths. they are all using urls.

library/bigip_software.py
caphrim007: see line 81

library/bigip_software.py
caphrim007: no you wouldnt, the url can just include those. you dont do any special parsing of the url

library/bigip_software.py
wojtek0806: they should include examples from your old module, unless i forgot to paste them

library/bigip_software.py
wojtek0806: ok i missed that bit in curl MAN page, i thought i have to specify this as curl parameters, i guess we can have basic auth

library/bigip_software.py
wojtek0806: but i think i will add to raise exception if: source.scheme is missing, as this is the only place i can see this breaking :)

library/bigip_software.py
wojtek0806: looks like i moved the metadata when copy paste :( doh!

library/bigip_software.py
wojtek0806: the local source examples are there, github has hidden them from you: 

https://github.com/F5Networks/f5-ansible/pull/249/files/c88d43e75db86dbac698a193c53d399657f92eaa#diff-2d56a17f77aa477d916dddac4da2ae3dL115

library/bigip_software.py
wojtek0806: the example already exists:

https://github.com/F5Networks/f5-ansible/pull/249/files/c88d43e75db86dbac698a193c53d399657f92eaa#diff-2d56a17f77aa477d916dddac4da2ae3dL206

wojtek0806: https://github.com/F5Networks/f5-ansible/pull/249/files/c88d43e75db86dbac698a193c53d399657f92eaa#diff-2d56a17f77aa477d916dddac4da2ae3dR294

library/bigip_software.py
wojtek0806: i have done it now, but the examples will not show this internally :) as we will not return changed filename as a return variable. I was thinking of doing that to inform user if the name changed, it is just a question of adding software_name and hotfix_name to the returned list.

library/bigip_software.py
wojtek0806: umm, you mean do this:

```
        if volume is not None:
            if state in ['activated', 'installed']:
                if self._volume_exists(volume):
                    self._check_active_volume(volume)
                else:
                    self._values['create_volume'] = True
                
                return self._values['volume']
```

library/bigip_software.py
wojtek0806: no it is not redundant, if the volumes list is 1 then we perform volume name increment on Active volume only. If the list is >1 then we just return next non-active volume combining this will break this logic and mess it up

library/bigip_software.py
wojtek0806: what exactly fails? if none of the statements are true then we return False, or return None, and in the code later we check if exists() is 'Truthy' this means that False or None values will fail that check. I do not understand what is the issue here ?

library/bigip_software.py
wojtek0806: again why? we can return None and in this case it will be fine as per my comment above

library/bigip_software.py
wojtek0806: have done the change but this leads to problems like this:

````````
Error
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/usr/local/lib/python2.7/dist-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/vagrant/ansible_test/test/unit/bigip/test_bigip_software.py", line 271, in test_activate_installed_volume
    results = mm.exec_module()
  File "/home/vagrant/ansible_test/library/bigip_software.py", line 805, in exec_module
    changed = self.activated()
  File "/home/vagrant/ansible_test/library/bigip_software.py", line 845, in activated
    return self.activate()
  File "/home/vagrant/ansible_test/library/bigip_software.py", line 970, in activate
    self._update_changed_options()
  File "/home/vagrant/ansible_test/library/bigip_software.py", line 833, in _update_changed_options
    attr2 = getattr(self.have, key)
  File "/home/vagrant/ansible_test/library/bigip_software.py", line 624, in volume
    raise F5ModuleError('You must specify a volume.')
F5ModuleError: You must specify a volume.

````````

This is what I have changed it to:

````````
    @property
    def volume(self):
        state = self._values['state']
        volume = self._values['volume']
        reuse = self._values['reuse_inactive_volume']

        if state in ['absent', 'present']:
            return self._values['volume']

        if volume is not None:
            if not self._volume_exists_on_device(volume):
                self._values['create_volume'] = True
            else:
                self._check_active_volume(volume)

            return self._values['volume']

        if volume is None:
            if reuse:
                new_volume = self._get_next_available_volume()
                if self._volume_exists_on_device(volume):
                    self._values['delete_volume'] = True
                    self._values['create_volume'] = True
                else:
                    self._values['create_volume'] = True
                self._values['volume'] = new_volume

                return self._values['volume']
            else:
                raise F5ModuleError('You must specify a volume.')

````````

I managed to overcome this bit by specifying api map for: {name: volume}, this is cause the returned volume json, uses 'name' parameter as volume ID.

However this leads to another issue, now i get this stacktrace on the same test:

``````

Error
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/usr/local/lib/python2.7/dist-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/vagrant/ansible_test/test/unit/bigip/test_bigip_software.py", line 271, in test_activate_installed_volume
    results = mm.exec_module()
  File "/home/vagrant/ansible_test/library/bigip_software.py", line 815, in exec_module
    changes = self.changes.to_return()
  File "/home/vagrant/ansible_test/library/bigip_software.py", line 751, in to_return
    result[returnable] = getattr(self, returnable)
  File "/home/vagrant/ansible_test/library/bigip_software.py", line 624, in volume
    raise F5ModuleError('You must specify a volume.')
F5ModuleError: You must specify a volume.

``````
We may have discovered a bug with this change, but I need your help to fix this

**Comment by wojtek0806** [comment]
changes made, had to iron out few bugs in the process, please review and retest this at your end @caphrim007

File: library/bigip_software.py (hunk 29 above)
Comment (caphrim007) [inline library/bigip_software.py]: you dont have a false to the true tests in those blocks.

```
        if software_path and hotfix_path:
            if self.image_exists_on_device() and self.hotfix_exists_on_device():
                return True
```

where is the case handled where if the first line is true, but the second line is false. The code would fall through and return None instead of False.

wojtek0806:
Ok then

library/bigip_software.py
wojtek0806: I know that will be None, but what I was looking at is HOW we use this function, and we only do:
if exists(): , which will evaluate to False if the return is either False or None. But if you want this for consistency that is fine.

Issue #250, opened 2017-06-22
IPv6 Addresses don't get compressed, input could use validation


champain:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME
`bigip_selfip`

##### ANSIBLE VERSION

```
ansible --version
ansible 2.3.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
  python version = 2.7.12 (default, Jul 25 2016, 16:14:54) [GCC 4.2.1 20070719 ]

```

##### PYTHON VERSION

```
Python 2.7.12
```

##### BIGIP VERSION

```
Sys::Version
Main Package
  Product     BIG-IP
  Version     12.1.2
  Build       0.0.249
  Edition     Final
  Date        Wed Nov 30 16:04:00 PST 2016

```

##### SUMMARY

IPv6 addresses can be entered into this module with or without leading 0s added. i.e. the address can be entered into the module as 
```
- name: Provision self-IPs
      bigip_selfip:
        address: fe80:500:110:727:0:0:0:11
        name: ipv6address
        netmask: 64
        vlan: v6vlan
        password: admin
        server: f5.bigip.local
        user: admin
        validate_certs: "no"
        allow_service: "none"
      delegate_to: localhost
```

But then the API will report the address back, stripped of all leading 0s

```
>>>ip = api.tm.net.selfips.selfip.load(name='v6-fe80-500-110-727-11', partition='Common')
>>>ip.attrs
{u'kind': u'tm:net:self:selfstate', u'partition': u'Common', u'name': u'v6-fe80-500-110-727-11', u'generation': 1291, u'addressSource': u'from-user', u'vlan': u'/Common/prov_external', u'trafficGroup': u'/Common/traffic-group-local-only', u'floating': u'disabled', u'selfLink': u'https://localhost/mgmt/tm/net/self/~Common~v6-fe80-500-110-727-11?ver=12.1.2', u'address': u'fe80:500:110:727::11/64', u'trafficGroupReference': {u'link': u'https://localhost/mgmt/tm/cm/traffic-group/~Common~traffic-group-local-only?ver=12.1.2'}, u'vlanReference': {u'link': u'https://localhost/mgmt/tm/net/vlan/~Common~prov_external?ver=12.1.2'}, u'fullPath': u'/Common/v6-fe80-500-110-727-11', u'unit': 0, u'inheritedTrafficGroup': u'false'} 
```

[...]

PR #251, opened 2017-06-22
adding validation to the IP address param and compressing for IPv6 checking

champain:
I'm not sure I did this right, but this is in response to https://github.com/F5Networks/f5-ansible/issues/250

### DIFF
--- a/library/bigip_selfip.py
+++ b/library/bigip_selfip.py
@@ -243,7 +243,7 @@ except ImportError:
     HAS_F5SDK = False
 
 try:
-    from netaddr import IPNetwork, AddrFormatError
+    from netaddr import IPNetwork, AddrFormatError, IPAddress
     HAS_NETADDR = True
 except ImportError:
     HAS_NETADDR = False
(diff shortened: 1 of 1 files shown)


caphrim007:
@champain,

Thanks for your contribution.
To ensure that the project team has proper rights to use your work, please complete the Contribution License Agreement at https://f5-ansible.readthedocs.io/en/devel/development/cla-landing.html.

It will cover your contributions to all F5-managed open source projects.
Thanks

champain:
I got signoff on this recently, so I should be able to get this to you shortly.

champain:
Alcon,

I've signed and sent the CLA to ansible_cla@f5.com. Let me know if you need anything else to move forward.

champain:
So the specific fix works in my testing environment when I patch it into ansible 2.2. However, the bigip_selfip module appears to be broken more generally within ansible 2.3, and I can't seem to be able to use it at all there. This is a tough position because you likely have little control over what version of ansible your module gets pulled into, but it puts me in a tough position because I have to patch this module, and then load it into my ansible environment externally. I honestly wish I had the bandwidth to submit a more full-featured pull request, because I think there is just an import path issue preventing this from working in some of the bigip modules and ansible 2.3 since I can see they've changed a lot of the directory structure.

For what it's worth here is the error I'm hitting when I try to load all of the master branch of the f5-ansible module in via a library folder

```
TASK [Provision self-IPs] **********************************************************************************************************************************************************************************
Wednesday 28 June 2017  14:58:46 -0400 (0:00:03.390)       0:00:03.390 ********
failed: [mlb3.proto.arin.net -> dj.pssid.cha.arin.net] (item={'key': u'v6-2001-500-110-727-11', 'value': {u'netmask': u'ffff:ffff:ffff:ffff::', u'vlan': u'prov_external', u'traffic_group': u'traffic-group
-local-only', u'address': u'2001:500:110:727:0::11'}}) => {"failed": true, "item": {"key": "v6-2001-500-110-727-11", "value": {"address": "2001:500:110:727:0::11", "netmask": "ffff:ffff:ffff:ffff::", "tra
ffic_group": "traffic-group-local-only", "vlan": "prov_external"}}, "msg": "The python f5-sdk module is required"}
```

caphrim007:
@champain the modules in this directory are always developed against the latest HEAD of ansible. we don't backport changes into previous releases.

This repo here, as mentioned in the docs and the readme, is not a stable repo, and the recommendation is to upgrade ansible if one wants a module that is in a newer version of ansible.

So if it works in the current devel branch (which will be 2.4) then that meets my expectations.

Issue #287, opened 2017-07-26
bigip_software_facts


wojtek0806:
- list of images and hotfixes present
- list of volumes installed along with software information

PR #294, opened 2017-08-01
Fixes #293

### LINKED ISSUE #293: add bigip_software_facts module
we need a new module for bigp_software_facts to be able to gather information about existing volumes and ISO images on the bigip

wojtek0806:
Problem:

New module for fetching bigip software facts was required

Analysis:

New module to gather volumes, images and hotfix information has been
added

Tests:
Unit
Functional

### DIFF
--- a/library/bigip_software_facts.py
+++ b/library/bigip_software_facts.py
@@ -75,12 +75,12 @@ images:
     type: list of dict
     sample:
         images:
-            - build: "0.0.184"
-              fileSize: "1997 MB",
-              lastModified: "Sun Oct  2 20:50:04 2016",
-              name: "BIGIP-12.1.1.0.0.184.iso",
-              product: "BIG-IP",
-              version: "12.1.1"
+            - build: 0.0.184
+              fileSize: 1997 MB,
+              lastModified: Sun Oct  2 20:50:04 2016,
+              name: BIGIP-12.1.1.0.0.184.iso,
+              product: BIG-IP,
+              version: 12.1.1
 hotfixes:
     description:
         List of hotfix ISOs that are present on the unit.
@@ -88,12 +88,12 @@ hotfixes:
     type: list of dict
     sample:
         hotfixes:
-            - build: "2.0.204"
-              fileSize: "1997 MB",
-              lastModified: "Sun Oct  2 20:50:04 2016",
-              name: "12.1.1-hf2.iso",
-              product: "BIG-IP",
-              version: "12.1.1"   
+            - build: 2.0.204
+              fileSize: 1997 MB,
+              lastModified: Sun Oct  2 20:50:04 2016,
+              name: 12.1.1-hf2.iso,
+              product: BIG-IP,
+              version: 12.1.1   
 volumes:     
     description:
         List the volumes present on device.
@@ -101,12 +101,12 @@ volumes:
     type: list of dict
     sample:
        volumes:
-            - basebuild: "0.0.184",
-              build: "0.0.184",
-              name: "HD1.2",
-              product: "BIG-IP",
-              status: "complete",
-              version: "12.1.1"              
+            - basebuild: 0.0.184,
+              build: 0.0.184,
+              name: HD1.2,
+              product: BIG-IP,
+              status: complete,
+              version: 12.1.1              
 '''
 
 
@@ -115,6 +115,7 @@ from ansible.module_utils.f5_utils import (
     AnsibleF5Parameters,
     HAS_F5SDK,
     F5ModuleError,
+    iteritems
 )
 
 
@@ -123,7 +124,7 @@ class Parameters(AnsibleF5Parameters):
         'name', 'build', 'version', 'product',
         'lastModified', 'basebuild', 'status',
         'active', 'fileSize'
-        ]
+    ]
 
     @property
     def include(self):
@@ -166,11 +167,37 @@ class Parameters(AnsibleF5Parameters):
 class ModuleManager(object):
     def __init__(self, client):
         self.client = client
+        self.want = Parameters(self.client.module.params)
+        self.include = self.want.include
 
     def exec_module(self):
-        manager = BaseManager(self.client)
+        if 'all' in self.include:
+            names = ['image', 'hotfix', 'volume']
+        else:
+            names = self.include
+        managers = [self.get_manager(name) for name in names]
+        result = self.execute_managers(managers)
+        return result
 
-        return manager.exec_module()
+    def execute_managers(self, managers):
+        results = dict(changed=False)
+        for manager in managers:
+            result = manager.exec_module()
+            for k, v in iteritems(result):
+                if k == 'changed':
+                    if v is True:
+                        results['changed'] = True
+                else:
+                    results[k] = v
+        return results
+
+    def get_manager(self, which):
+        if 'image' == which:
+            return ImageFactManager(self.client)
+        if 'hotfix' == which:
+            return HotfixFactManager(self.client)
+        if 'volume' == which:
+            return VolumeFactManager(self.client)
 
 
 class BaseManager(object):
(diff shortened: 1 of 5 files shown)


wojtek0806:
this is related to #287

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+short_description: Collect software facts from BIG-IP devices.
+description:
+  - Collect information about installed volumes, existing ISOs for images and hotfixes on the BIG-IP device.
+version_added: "2.5"
+options:
+  include:
+    description:
+      - Type of information to collect.
+    required: False
+    default: ['all']
```
caphrim007: this should be a "-" list like the `choices` option shown below

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+module: bigip_software_facts
+short_description: Collect software facts from BIG-IP devices.
+description:
+  - Collect information about installed volumes, existing ISOs for images and hotfixes on the BIG-IP device.
+version_added: "2.5"
+options:
+  include:
+    description:
+      - Type of information to collect.
+    required: False
```
caphrim007: Remove `required: False` because this is the default

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+from ansible.module_utils.f5_utils import (
+    AnsibleF5Client,
+    AnsibleF5Parameters,
+    HAS_F5SDK,
+    F5ModuleError,
+)
+
+
+class Parameters(AnsibleF5Parameters):
+    returnables = ['name', 'build', 'version',
```
caphrim007: Please start this list on a newline. And end the list so that it is aligned with the variable name.
ex.
```
var1 = [
    'foo'
]
```

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+            raise F5ModuleError(error)
+
+        if key not in keys:
+            raise F5ModuleError('"{0}" is not a supported filter. '
+                                'Supported key values are: {1}'.format(key, ', '.join(keys)))
+
+        return key, value
+
+
+class FactManagerBase(object):
```
caphrim007: This should be called `BaseManager`

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+    def __init__(self, client):
+        self.client = client
+        self.want = Parameters(self.client.module.params)
+        self.result = dict()
+        self.include = self.want.include
+        self.filter = self.want.filter
+
+    def display_facts(self):
```
caphrim007: this should be `exec_module`

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+            for item in collection:
+                if hasattr(item, key):
+                    if getattr(item, key) == value:
+                        output.append(self._filter_and_format_facts(item))
+        return output
+
+
+class ImageFactManager(FactManagerBase):
+
+    def get_images_facts(self):
```
caphrim007: let's standardize on this method name and call it `get_facts`

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+        collection = self.list_images_on_device()
+        to_return = self.collection_parser(collection)
+        return to_return
+
+    def list_images_on_device(self):
```
caphrim007: let's standardize on this method name and call it `get_facts_from_device`

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+    spec = ArgumentSpec()
+
+    client = AnsibleF5Client(
+        argument_spec=spec.argument_spec,
+        supports_check_mode=spec.supports_check_mode,
+        f5_product_name=spec.f5_product_name
+    )
+
+    try:
+        mm = FactManagerBase(client)
```
caphrim007: We need to change this so that `ModuleManager` is used instead of the base manager. Only manager can be used in the main method

library/bigip_software_facts.py
```diff
@@ -0,0 +1,306 @@
...
+        results = mm.display_facts()
```
caphrim007: this should be `exec_module`

test/integration/targets/bigip_software_facts/tasks/main.yaml
```diff
@@ -0,0 +1,76 @@
+---
+- name: Get all software facts
+  bigip_software_facts:
+  register: result
+
+- name: Assert Get all software facts
+  assert:
+      that:
+          - result|changed
+          - "'images' in result"
```
caphrim007: good job doing this!

test/unit/bigip/test_bigip_software_facts.py
```diff
@@ -0,0 +1,389 @@
...
+from ansible.compat.tests import unittest
+from mock import patch
+from mock import Mock
+from ansible.module_utils import basic
+from ansible.module_utils._text import to_bytes
+from ansible.module_utils.f5_utils import AnsibleF5Client
+
+try:
+    from library.bigip_software_facts import Parameters
+    from library.bigip_software_facts import FactManagerBase
```
caphrim007: This needs to be ModuleManager after you refactor the Base manager

test/unit/bigip/test_bigip_software_facts.py
```diff
@@ -0,0 +1,389 @@
...
+    from library.bigip_software_facts import ArgumentSpec
+    from library.bigip_software_facts import F5ModuleError
+except ImportError:
+    try:
+        from ansible.modules.network.f5.bigip_software_facts import Parameters
+        from ansible.modules.network.f5.bigip_software_facts import FactManagerBase
```
caphrim007: ModuleManager

test/unit/bigip/test_bigip_software_facts.py
```diff
@@ -0,0 +1,389 @@
...
+            f5_product_name=self.spec.f5_product_name
+        )
+
+        mm = FactManagerBase(client)
+        mm.exit_json = Mock(return_value=False)
+        results = mm.display_facts()
+
+        assert results['changed'] is True
+        assert len(results['hotfixes']) == 0
+        assert results['images'] == [{'product': 'BIG-IP', 'name': 'BIGIP-12.1.1.0.0.184.iso',
```
caphrim007: it's a bad idea to assert the structure of a dictionary because dicts are by nature, unstructured. Instead, assert the following

* len of the list == 1
* that the product key exists in [0]
* that the name key exists in [0]
* that the value of [0]['product'] is 'BIG-IP'
* that the value of [0]['product'] is 'BIGIP-12.1.1.0.0.184.iso'

Do this change for _all_ dict assertions

test/unit/bigip/test_bigip_software_facts.py
```diff
@@ -0,0 +1,389 @@
...
+            argument_spec=self.spec.argument_spec,
+            supports_check_mode=self.spec.supports_check_mode,
+            f5_product_name=self.spec.f5_product_name
+        )
+
+        with pytest.raises(F5ModuleError) as err:
+            msg = 'Include parameter may only be specified as one or more of the following: all, volume, image, hotfix'
+            FactManagerBase(client)
+
+        assert err.value.message == msg
```
caphrim007: to be python 3 compatible, all of these err.value.message assertions need to be `str(err.value)`

wojtek0806:
requested changes pushed :) @caphrim007

wojtek0806:
Any progress on this? @caphrim007

library/bigip_software_facts.py
```diff
@@ -0,0 +1,317 @@
...
+RETURN = '''
+images:
+    description:
+        List of base image ISOs that are present on the unit.
+    returned: changed
+    type: list of dict
+    sample:
+        images:
+            - build: "0.0.184"
+              fileSize: "1997 MB",
```
caphrim007: Remove commas, they're not necessary

library/bigip_software_facts.py
```diff
@@ -0,0 +1,317 @@
...
+    F5ModuleError,
+)
+
+
+class Parameters(AnsibleF5Parameters):
+    returnables = [
+        'name', 'build', 'version', 'product',
+        'lastModified', 'basebuild', 'status',
+        'active', 'fileSize'
+        ]
```
caphrim007: this bracket aligns under the `returnables` variable, backspace 4 spaces

library/bigip_software_facts.py
```diff
@@ -0,0 +1,317 @@
...
+
+        return key, value
+
+
+class ModuleManager(object):
+    def __init__(self, client):
+        self.client = client
+
+    def exec_module(self):
+        manager = BaseManager(self.client)
```
caphrim007: I think we missed the idea here. The base manager is what gets inherited by the concrete classes, it's not what is actually run.

What gets run should be the module manager which does the logic of which filter to use. So this Manager should really look more like the bigip_ssl_certificate ModuleManager where it gets a variety of managers and combines their results.

library/bigip_software_facts.py
```diff
@@ -0,0 +1,317 @@
...
+        if 'all' in self.include:
+            facts = self.get_all_facts()
+        else:
+            facts = self.get_selected_facts()
+
+        result.update(**facts)
+        result.update(dict(changed=True))
+        return result
+
+    def get_all_facts(self):
```
caphrim007: something like this method, but not this method, should live in the module manager. There is no need to list all the managers here. Instead, with the get_managers() method that you'll find in bigip_ssl_certificate, you should be able to call all the get_facts on their own

caphrim007:
lgtm

PR #296, opened 2017-08-05
Update bigip_selfip.py

Blaag:
Default traffic group should probably come from /Common - there does not appear to be any way to explicitly call traffic groups from /Common if you are in a different partition.

### DIFF
--- a/library/bigip_selfip.py
+++ b/library/bigip_selfip.py
@@ -593,7 +593,7 @@ class BigIpSelfIp(object):
             )
 
         if traffic_group is None:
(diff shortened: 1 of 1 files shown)


caphrim007:
@Blaag the way to call traffic groups from /Common if you are in a different partition is to name them explicitly.

ex.
```
bigip_selfip:
    partition: foo
    traffic_group: "/Common/traffic-group-1
```

While I do appreciate the effort to send this patch, a refactored selfip module was just pushed up here which should include the aforementioned syntax

Blaag:
Thanks much for the feedback, will check out the new module.

Appreciate it!

PR #305, opened 2017-08-10
netaddr mistakenly given IP address including route domain, e.g. "1.2…

Blaag:
….3.4%0"

When checking for a valid destination address on a virtual server, netaddr always fails because it is given an address in the format of "1.2.3.4%0" where there route domain is being included.

### DIFF
--- a/library/bigip_virtual_server.py
+++ b/library/bigip_virtual_server.py
@@ -229,7 +229,7 @@ class Parameters(AnsibleF5Parameters):
             return None
         destination = self._values['destination']
         try:
-            netaddr.IPAddress(destination)
+            netaddr.IPAddress(destination.split("%")[0])
         except netaddr.core.AddrFormatError:
             raise F5ModuleError(
                 "The provided destination is not a valid IP address"


caphrim007:
@Blaag ,

Thanks for your contribution.
To ensure that the project team has proper rights to use your work, please complete the Contribution License Agreement at https://f5-ansible.readthedocs.io/en/devel/development/cla-landing.html.

It will cover your contributions to all F5-managed open source projects.
Thanks

Blaag:
Awesome, just sent my completed CLA.

Thanks!

caphrim007:
@Blaag do route domains work on the virtual_server module that currently ships with ansible?

Blaag:
I hadn't checked that, but yes the same play against default install of ansible 2.3.1.0 does work.

caphrim007:
@Blaag the reason I ask is because in other modules we've taken the approach of one specifying the route domain in a `route_domain` argument instead of as part of the address itself.

So that's the longer term approach that I would prefer. I think in the interim, to support deprecation, we can do a play on this PR. I'll merge this, but we'll need to change it later to support both cases

Blaag:
Makes sense. Thanks for the feedback!

PR #319, opened 2017-08-30
Add ability to assign trunks to vlans

nahun:
This adds trunks to the list of interfaces that can be assigned to a VLAN

### DIFF
--- a/library/bigip_provision.py
+++ b/library/bigip_provision.py
@@ -51,6 +51,7 @@ options:
       - pem
       - sam
       - swg
+      - vcmp
   level:
     description:
       - Sets the provisioning level for the requested modules. Changing the
@@ -298,7 +299,7 @@ class ArgumentSpec(object):
                 choices=[
                     'afm', 'am', 'sam', 'asm', 'avr', 'fps',
                     'gtm', 'lc', 'ltm', 'pem', 'swg', 'ilx',
-                    'apm'
+                    'apm', 'vcmp'
                 ]
             ),
             level=dict(
--- a/library/bigip_vlan.py
+++ b/library/bigip_vlan.py
@@ -246,6 +246,7 @@ class Parameters(AnsibleF5Parameters):
 
     def _get_interfaces_from_device(self):
         lst = self.client.api.tm.net.interfaces.get_collection()
+        lst.append(self.client.api.tm.net.trunks.get_collection())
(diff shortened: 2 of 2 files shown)


caphrim007:
@nahun I dont think `append` is what you want. I think `extend` is

nahun:
You're right, I didn't see that since its actually working for me. The error isn't being raised correctly it seems. If I define a non-existent interface its the BIG-IP that throws the error:
```
failed: [hostname -> localhost] (item={u'tag': 2223, u'name': u'test', u'tagged_interfaces': [u'1/1.17']}) => {"failed": true, "item": {"name": "test", "tag": 2223, "tagged_interfaces": ["1/1.17"]}, "msg": "400 Unexpected Error: Bad Request for uri: https://1.1.1.1:443/mgmt/tm/net/vlan/~Common~test/\nText: u'{\"code\":400,\"message\":\"01070277:3: The requested vlan member (1/1.17) was not found.\",\"errorStack\":[]}'"}
```
And then trying the _parse_return_ifcs method manually with the SDK throws an attribute error as I would expect, but the method doesn't fail when running a playbook.
```
>>> from f5.bigip import ManagementRoot
>>> mgmt = ManagementRoot("1.1.1.1", "admin", "admin")
>>> l = mgmt.tm.net.interfaces.get_collection()
>>> l.append(mgmt.tm.net.trunks.get_collection())
>>> t = [str(x.name) for x in l]
  File "<stdin>", line 1, in <module>
AttributeError: 'list' object has no attribute 'name'
```

caphrim007:
the bigger question is why didn't you get an exception using the trunks API.

because there is no trunks API

caphrim007:
what version of python are you using

nahun:
Oh man I completely forgot I added that myself as well. Alright next time I won't put in a pull request months after I do some work.

In the F5 SDK I added a simple bigip/tm/net/trunk.py that is pretty much copied from interface.py. Maybe I'll contribute it to f5-common-python.

Closing this since there isn't even support for it in the SDK. Sorry about that.

caphrim007:
well, I added and pushed the vcmp addition you made. so that should now be in the provision code

nahun:
ok thanks, I also didn't mean for that to be in this request. I was going to open a different one. I of course should have created different branches for those...

PR #339, opened 2017-09-25
BIGIP_POOL: Update module to support 'single' monitor_type and fix id…

jmcgill298:
…empotency issue

### DIFF
--- a/library/bigip_pool.py
+++ b/library/bigip_pool.py
@@ -73,7 +73,7 @@ options:
       - Monitor rule type when C(monitors) > 1. When creating a new pool, if this
         value is not specified, the default of 'and_list' will be used.
     version_added: "1.3"
-    choices: ['and_list', 'm_of_n']
+    choices: ['and_list', 'm_of_n', 'single']
   quorum:
     description:
       - Monitor quorum value when C(monitor_type) is C(m_of_n).
@@ -265,9 +265,8 @@ class Parameters(AnsibleF5Parameters):
     }
 
     updatables = [
-        'monitor_type', 'quorum', 'monitors', 'service_down_action',
-        'description', 'lb_method', 'slow_ramp_time', 'reselect_tries',
-        'host', 'port'
+        'quorum', 'monitors', 'service_down_action', 'description',
+        'lb_method', 'slow_ramp_time', 'reselect_tries', 'host', 'port'
     ]
 
     returnables = [
@@ -371,11 +370,11 @@ class Parameters(AnsibleF5Parameters):
         if self._values['monitors'] is None:
             return None
         monitors = [self._fqdn_name(x) for x in self.monitors_list]
-        if self.monitor_type == 'and_list':
-            result = ' and '.join(monitors).strip()
-        else:
+        if self.monitor_type == 'm_of_n':
             monitors = ' '.join(monitors)
             result = 'min %s of { %s }' % (self.quorum, monitors)
+        else:
+            result = ' and '.join(monitors).strip()            
         return result
 
     @property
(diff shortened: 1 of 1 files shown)


jmcgill298:
```
ansible version 2.4.0.0
python version = 2.7.12 (default, Nov 19 2016, 06:48:10) [GCC 5.4.0 20160609]
F5 versoin = BIG-IP 12.1.1 Build 0.0.184
```

The F5 supported monitor types are: none, single, and_list, and m_of_n. This PR adds support for single type. 

Additionally, the F5 Rest API does not return a value for monitor_type. This seems to be an attribute derived from the number of monitors used and the quorum value, not something that is configurable. Since it is not returned with the existing data, the module reports it as updated every time. I removed monitor_type from updatables to fix this.

caphrim007:
@jmcgill298 can you provide an example YAML that illustrates the idempotency problem?

caphrim007:
I think you've interpreted the updateables array incorrectly. The updatables defines what the module will consider to be a parameter that can be updated; it's not a one-to-one translation of the API resource's attributes.

monitor_type must continue to exist because it is the only way that people can specify if they want to use "all", or, "at least" the list of monitors. There may indeed be an idempotency issue, but excluding monitor_type is not the solution to its fix.

YAML that repros the issue will be helpful to have.

jmcgill298:
```
- name: Modify load balancer method
  bigip_pool:
      server: "lb.mydomain.com"
      user: "admin"
      password: "secret"
      state: "present"
      name: "my-pool"
      partition: "Common"
      monitors:
      - "http"
      - "tcp"
      monitor_type: "and_list"
```

Task always returns as changed and monitor_type is what is changed. Monitor type is not returned by the API when I take a GET in Postman

jmcgill298:
I think you misunderstood my point about it not being returned by the API. The point was to simply say that it will always be None when you do `getattr(self.have, 'monitor_type')`, which means the module will try to update the monitor type whenever `monitor_type` is defined.

I thought it made sense to remove `monitor_type` from `updatables` since it is not a configurable property from the perspective of the F5, but a descriptive attribute. Maybe that is wrong.

caphrim007:
ahh ok. In that case, it's not that monitor_type is wrong to have in the module, it's that it is not being set correctly in the module to begin with. We can certainly fix this. Is `single` still a choice that you want to use? We would alias it to `and_list` but it may provide a more intuitive usage for folks that think a single monitor should be a "single" type.

jmcgill298:
It is definitely not needed, but I thought it would be more intuitive for users. My use case is to use the get_facts module, use this data to build the VIP config, then push the config back using the necessary modules.

The SOAP API does return the monitor type attribute, so I built a mapper to use for model param values and assumed single type was single. So, I would say that "single" type seems more natural. It would also make sense to have if later you wanted to do data validation between monitor_type, monitors, and quorum

caphrim007:
@jmcgill298 ok, that's a reasonable request. I want to close this PR and re-open it as an issue to fix the idempotency and add the 'single' choice if that's ok with you.

jmcgill298:
That is fine with me. When do you think you will have a fix? Is what you have here going to be upstreamed to Ansible core?

caphrim007:
@jmcgill298 I can add the fix tomorrow. Everything in this repo eventually makes it upstream, we were required to hold off for a period of time because Ansible's merge window was closed during the 2.4 release.

The 2.5 work we're looking to accomplish is listed here
https://github.com/F5Networks/f5-ansible/projects/2

jmcgill298:
thanks @caphrim007

jmcgill298:
#339

jmcgill298:
#341

caphrim007:
@jmcgill298 closing this as the aforementioned Issues should now be fixed. Also, in the future, if you want to submit another PR, we have a CLA that F5 legal wants contributors to submit. Details on that can be found here https://f5-ansible.readthedocs.io/en/devel/development/cla-landing.html

PR #351, opened 2017-09-27
fix typo;add undo example

chen23:
the undo example is not the most well thought out, the typo is a current issue.


caphrim007:
lgtm

PR #356, opened 2017-10-04
add virtualenv; soften not using pip; example of using bleeding-edge

chen23:
A customer was trying to install Ansible on their Mac and failing.  These changes are an attempt to provide a sane starting point for someone to use F5 Ansible and knows nothing about either Ansible and/or Python.

Some of this content "borrowed" from @mcgonagle https://github.com/mcgonagle/ansible_f5/blob/master/docs/INSTALL.md

### DIFF
--- a/docs/make.bat
+++ b/docs/make.bat
@@ -1,0 +1,281 @@
+@ECHO OFF
+
+REM Command file for Sphinx documentation
+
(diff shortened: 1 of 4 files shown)


caphrim007:
@chen23 this is fine

PR #366, opened 2017-10-12
Asm policy import activate

wojtek0806:
I still have to implement functional tests, but this PR will allow importing and management of ASM policies


library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+    @property
+    def template(self):
+        tmpl_name = self._values['template']
+        if tmpl_name is None:
+            return None
+        if self._template_exists_on_device(tmpl_name):
+            return self._values['template']
+        else:
+            raise F5ModuleError('Template with the given name: {0} does not exist'.format(tmpl_name))
+
```
caphrim007: i generally do not attempt to check to see if something already exists like this because MCPD will do this for you. You're either making a round-trip to bigip here, or you're making a round-trip to bigip during the update (when mcpd will catch your error). Therefore I delegate that to mcpd because I am not the expert.

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+    def _template_exists_on_device(self, name):
+        collection = self._templates_on_device()
+        for resource in collection:
+            if resource.name == name.upper():
+                self._set_template_selflink(resource)
+                return True
+        return False
+
+    def _set_template_selflink(self, template):
```
caphrim007: not needed if `_template_exists_on_device` is remove

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+
+
+class ModuleManager(object):
+    def __init__(self, client):
+        self.client = client
+        self.have = None
+        self.want = Parameters()
+        self.want.client = self.client
+        self.want.update(self.client.module.params)
+        self.changes = Parameters()
```
caphrim007: change this to `Changes()` as we use a changes class to hold these changes. This additionally helps make callgraphs easier to read

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+        result.update(dict(changed=changed))
+        return result
+
+    def _set_changed_options(self):
+        changed = {}
+        for key in Parameters.returnables:
+            if getattr(self.want, key) is not None:
+                changed[key] = getattr(self.want, key)
+        if changed:
+            self.changes = Parameters(changed)
```
caphrim007: change this to `Changes(changed)`

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+    def _update_changed_options(self):
+        changed = {}
+        for key in Parameters.updatables:
+            if getattr(self.want, key) is not None:
+                attr1 = getattr(self.want, key)
+                attr2 = getattr(self.have, key)
+                if attr1 != attr2:
+                    changed[key] = attr1
+        if changed:
+            self.changes = Parameters(changed)
```
caphrim007: change this to `Changes(changed)`

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+        if self.exists():
+            return self.update()
+        else:
+            return self.create()
+
+    def absent(self):
+        if not self.exists():
+            return False
+        else:
+            if self.client.check_mode:
```
caphrim007: move this to the `delete()` method

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+                return True
+            self.delete()
```
caphrim007: `delete()` method itself should be returning the return value. The `return True` below should be part of the `delete()` method

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+            return True
+
+    def exists(self):
+        if self.client.check_mode:
```
caphrim007: check mode does not need to be in an `exists` method. check mode is used when things have the potential for changing.

test/unit/bigip/test_bigip_asm_policy.py
```diff
@@ -0,0 +1,580 @@
...
+from ansible.compat.tests.mock import patch, Mock, PropertyMock
+from ansible.module_utils import basic
+from ansible.module_utils._text import to_bytes
+from ansible.module_utils.f5_utils import AnsibleF5Client
+from ansible.module_utils.f5_utils import F5ModuleError
+
+try:
+    from library.bigip_asm_policy import Parameters
+    from library.bigip_asm_policy import ModuleManager
+    from library.bigip_asm_policy import ArgumentSpec
```
caphrim007: iControlUnexpectedHTTPError needs to be imported here. see below for an example

https://github.com/F5Networks/f5-ansible/blob/devel/test/unit/bigip/test_bigip_software.py#L45

test/unit/bigip/test_bigip_asm_policy.py
```diff
@@ -0,0 +1,580 @@
...
+except ImportError:
+    try:
+        from ansible.modules.network.f5.bigip_asm_policy import Parameters
+        from ansible.modules.network.f5.bigip_asm_policy import ModuleManager
+        from ansible.modules.network.f5.bigip_asm_policy import ArgumentSpec
```
caphrim007: iControlUnexpectedHTTPError needs to be imported here. see below for an example

https://github.com/F5Networks/f5-ansible/blob/devel/test/unit/bigip/test_bigip_software.py#L45

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+        else:
+            return False
+
+    def return_policy(self):
+        policies = self.policies_on_device()
+        for policy in policies:
+            if policy.name == self.want.name:
+                return policy
+
+    def policy_exists_on_device(self):
```
caphrim007: this doesnt need to exist. This method can be combined into the `exists` method

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+            if self.want.active:
+                self.activate()
+                return True
+            else:
+                return True
+        else:
+            return False
+
+    def update(self):
+        policy = self.return_policy()
```
caphrim007: this method should be `self.read_current_from_device()`, and that should assign to self.have

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+        return result
+
+    def policy_active(self):
+        policy = self.return_policy()
+        if policy.active is True:
+            return True
+        else:
+            return False
+
+    def return_policy(self):
```
caphrim007: this should be named `read_current_from_device` and should look like the method shown here

https://github.com/F5Networks/f5-ansible/blob/devel/library/bigip_device_sshd.py#L256

library/bigip_asm_policy.py
```diff
@@ -0,0 +1,546 @@
...
+        return result
+
+    def create_policy_on_device(self):
+        result = self.client.api.tm.asm.policies_s.policy.create(name=self.want.name)
+        return result
+
+    def delete_policy_on_device(self):
+        policy = self.return_policy()
+        policy.delete()
+        if policy.exists():
```
caphrim007: this existence should be checked in the `remove()` method.

Also, this module is misnamed. it should be `remove_from_device`

library/bigip_asm_policy.py
wojtek0806: We need to be able to check for this so the create on device operstion does not fail

PR #397, opened 2017-10-26
added f5_utils import

tribeiros:
Added _utils missing


caphrim007:
@tribeiros we do not develop against the `master` branch. Additionally, this code has already been fixed. You can find it here

https://github.com/F5Networks/f5-ansible/blob/devel/library/_bigip_pool_member.py

PR #424, opened 2017-11-07
Fixed passive voice, standardized headings, reorganized

selhorn:
I ran some scripts to review for passive voice and fixed the related issues. I reorganized the order of the getting started to be: install, get support, run your first playbook, and then install unsupported modules.


caphrim007:
@selhorn conflicts exist in the PR. Will wait for them to be fixed

PR #428, opened 2017-11-08
Bugfix.formatting and standards

RobCupples:
Formatting and standards updates to config, configsync_action, device_connectivity, and device_dns modules.

- Decreased indents from 4 to 2 spaces
- Replaced '- include: *.yaml' with '-import_tasks: *.yaml'
- Replaced 'with_items:' with 'loop:'


caphrim007:
lgtm

PR #435, opened 2017-11-10
Coding standards: facts, gtm_datacenter, gtm_facts, gtm_pool

RobCupples:
Updating coding standards for facts, gtm_datacenter, gtm_facts, and gtm_pool modules.

1. 2-space indents
2. import_tasks: *.yaml instead of include *.yaml
3. loop: instead of with_items:


test/integration/targets/bigip_gtm_facts/tasks/main.yaml
```diff
@@ -4,115 +4,115 @@
...
+      - "'server' in result"
 
 - name: Get GTM pool facts, all specific
   bigip_gtm_facts:
-      include:
-        - pool
-        - wide_ip
-        - server
+    include:
+    - pool
```
caphrim007: these are not spaced over far enough

RobCupples:
My mistake.  I probably would have found this if I hadn't gotten lazy today and stopped running tests after updating them.  Fixed with an additional commit.  Ran all related modules.

caphrim007:
lgtm

PR #439, opened 2017-11-13
Formatting and deprecation updates for gtm_server, gtm_virtual_server, and gtm_wide_ip

RobCupples:
Converting to 2-space indents
Replace include: *.yaml with import_tasks: *.yaml
Replace with_items: with loop:


test/integration/targets/bigip_gtm_server/tasks/main.yaml
```diff
@@ -2,140 +2,140 @@
...
-      datacenter: "{{ dc_name1 }}"
-      state: "present"
+    name: "{{ server_name }}"
+    devices:
+      - name: "foo"
+        addresses:
+          - address: "4.4.4.1"
+            translation: "192.168.14.10"
+          - address: "4.4.4.2"
+        translation: "1.1.1.1"
```
caphrim007: You've identified a bug in the tests. This should actually go underneath the `address` above it.

test/integration/targets/bigip_gtm_server/tasks/main.yaml
```diff
@@ -2,140 +2,140 @@
...
```
caphrim007: Same bug I mentioned above applies here. The bug is duplicated because when I write idempotent tests, I deliberately copy and paste the original test.

RobCupples:
My environment is so screwed up I think I'm going to have to start over on a new branch.

PR #447, opened 2017-11-15
Standards in hostname, irule, and license modules.

RobCupples:
2-space indents.
Replace "include: *.yaml" with "import_tasks: *.yaml"


caphrim007:
lgtm

PR #451, opened 2017-11-15
Applying standards to node, partition, policy, and policy_rule modules.

RobCupples:
- Standardize on 2-space indents
- Replace "with_items:" with "loop:"
- Replace "include: *.yaml" with "import_tasks: *.yaml"


caphrim007:
lgtm

PR #459, opened 2017-11-17
Formating and standardization for provision, qkview, raw, and remote_…

RobCupples:
…syslog modules along with 1 missed update in pool_member module.

- standardize on 2-space indents
- replace 'include: *.yaml' with 'import_tasks: *.yaml'
- replace 'with_items:' with 'loop:'


caphrim007:
lgtm

PR #461, opened 2017-11-17
docs build, testing, and deployment

jputrino:
- Add docs test script to travis scripts section.
- Add docs deployment to makefile; add make command for deployment to travis.


caphrim007:
@jputrino do you want to remove that grammar check and put it in some other side-band process? @selhorn is ultimately responsible for that, but it's going to block many PRs before its fixed

PR #462, opened 2017-11-17
Formatting and standards to routedomain and routedomain_facts modules.

RobCupples:
- 2-space indents
- replace 'include: *.yaml' with 'import_tasks: *.yaml'
- replace 'with_items:' with 'loop:'


caphrim007:
lgtm

PR #476, opened 2017-11-29
Continuuing standards and formatting updates to integration tests.

RobCupples:
Updates to integration tests for bigip_view, bigip_vlan, bigip_wait, bigiq_license_pool,  bigiq_license_pool_member,  and f5_support_upload modules.  The view and vlan modules were previously updated but grep and random inspection revealed needed updates that were previously missed.

- Standardize on 2-space indents
- Replace 'include: *.yaml' with 'import_tasks: *.yaml'
- Replace 'with_items:' with 'loop:'

### DIFF
--- a/test/integration/bigip_vlan.yaml
+++ b/test/integration/bigip_vlan.yaml
@@ -44,11 +44,11 @@
         - ansible.module_utils.six.*
 
   environment:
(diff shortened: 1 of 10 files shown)


caphrim007:
lgtm

PR #485, opened 2017-12-04
Bugfix.docs edits development directory

selhorn:
Updated conf.py to remove reference to tests so that rst files including test in the name would build
Fixed style issues that were preventing code from being displayed
Fixed passive voice issues
Removed extra issues file from dev directory and put the info into usage\filing-issues file
First round of edits to make the content conform more closely to F5 style

### DIFF
--- a/docs/development/code-conventions.rst
+++ b/docs/development/code-conventions.rst
@@ -1,32 +1,18 @@
 Code Conventions
 ================
 
-The F5 modules attempt to follow a set of coding conventions that apply to
-all new and existing modules.
+The F5 modules attempt to follow a set of coding conventions that apply to all new and existing modules. These conventions help new contributors quickly develop new modules, and they help existing contributors maintain the current modules.
 
-These conventions help new contributors quickly develop new modules.
-Additionally, they help existing contributors maintain the current modules.
+Where possible, F5 tries to automate the validation of these coding conventions so you are aware of mistakes and can fix them yourself.
 
-Style checking
---------------
+For more information on the tools that perform these checks, refer to the :doc:`tests` page.
 
-Where possible, we try to automate the validation of these coding conventions
-so that you are aware of mistakes and able to fix them yourself without
-having to have the maintainers intervene.
-
-For more information on what tools perform these checks, refer to the tests
-page.
-
-Module Conventions
-------------------
-
-When writing your modules and their accompanying tests and docs, please
-follow the below coding conventions.
+When you write modules and their accompanying tests and docs, follow these coding conventions.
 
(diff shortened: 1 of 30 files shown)


docs/development/code-conventions.rst
```diff
@@ -1,32 +1,18 @@
...
-When writing your modules and their accompanying tests and docs, please
-follow the below coding conventions.
+When you write modules and their accompanying tests and docs, follow these coding conventions.
 
 Use the complex/structure map format
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+------------------------------------
 
-In reference to Jeff Geerling's page `here`_, this format looks like this.
+In reference to Jeff Geerling's page, this format looks like this.
```
caphrim007: This line contains a URL ref, this shouldn't be changed

docs/development/code-conventions.rst
```diff
@@ -516,27 +454,21 @@ allow you to debug the problem.
...
-operation. This is very handy when you want to run a set of tasks but
-are not sure what will happen when you do.
+Check mode allows Ansible to run your Playbooks in a dry-run sort of operation. This is handy when you want to run a set of tasks but are not sure what will happen when you do.
 
-Since BIG-IPs are usually considered a sensitive device to handle, there
-should always be a check-mode implemented in your module.
+Because BIG-IPs are usually considered a sensitive device to handle, you should always implement a check mode.
 
-.. _here: http://www.jeffgeerling.com/blog/yaml-best-practices-ansible-playbooks-tasks
+http://www.jeffgeerling.com/blog/yaml-best-practices-ansible-playbooks-tasks
```
caphrim007: this is the page that the ref was referring to. this needs to be reverted.

docs/development/code-conventions.rst
```diff
@@ -564,14 +496,14 @@ use `delegate_to`.
          key: "setup.run"
          state: "reset"
 
-Default EXAMPLE parameters
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+Format common parameters consistently
```
caphrim007: this header needs to be reverted. EXAMPLE is a literal block in the ansible module. So it is not sufficient to just say "format common parameters". We need to very specifically format the EXAMPLES parameter with the mentioned params.

selhorn: I am trying to make all the titles actions. Can I use something like "Set default EXAMPLE parameters"?

caphrim007: @selhorn that's fine

PR #497, opened 2017-12-08
more edits, passive fixes, updated test-docs to not grammar check mod…

selhorn:
More grammar and passive voice changes, tried to standardize title caps


selhorn:
Let me revert that "write-good" change that probably caused the error...

PR #514, opened 2017-12-13
Tests corresponding to issue 513

RobCupples:
Try to change link_discovery and virtual_server_discovery in bigip_gtm_server after the server has been created.

### DIFF
--- a/test/integration/targets/bigip_gtm_server/tasks/main.yaml
+++ b/test/integration/targets/bigip_gtm_server/tasks/main.yaml
@@ -141,3 +141,4 @@
     - "{{ dc_name2 }}"
 
 - import_tasks: issue-00513.yaml
+  tags: issue-00513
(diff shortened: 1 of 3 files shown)


test/integration/targets/bigip_gtm_server/tasks/main.yaml
```diff
@@ -139,3 +139,5 @@
   loop:
     - "{{ dc_name1 }}"
     - "{{ dc_name2 }}"
+
+- import_tasks: issue-00513.yaml
```
caphrim007: Add a `tags` line underneath this import so that we can test these out-of-band of the entire integration suite.

for example,

`tags: issue-00513`

RobCupples: Done.

RobCupples: Done.

PR #672, opened 2018-03-19
f5 guide draft

selhorn:
Draft content for Ansible scenarios guide

### DIFF
--- a/docs/usage/guide_f5.rst
+++ b/docs/usage/guide_f5.rst
@@ -1,0 +1,316 @@
+:orphan: true
+
+F5 BIG-IP Guide
+===============
+
+BIG-IP systems can inspect, secure, balance, and accelerate the traffic that passes through your network.
+
+F5 provides modules for Ansible that you can use to deploy, provision, configure, and orchestrate BIG-IP systems. BIG-IP can be physical hardware or a virtual edition (BIG-IP VE) running in a public or private cloud.
+
+About the API
+`````````````
+
+The primary API that's included with BIG-IP is called iControl REST. The Python library that interacts with iControl REST is called ``f5-sdk``.
+
+To work with the F5 Modules for Ansible, you should install ``f5-sdk``, as well as:
+
+- ``bigsuds``, the F5 SOAP API that was used for modules prior to Ansible 2.2
+- ``netaddr``, which helps with address manipulation
+- ``deepdiff``, to compare and find the differences between objects
+
+
+For details:
+
+- `Learn more about the f5-sdk Python library <http://f5-sdk.readthedocs.io/en/latest/userguide/index.html>`_
+
+- `Access the f5-sdk Python library <https://github.com/F5Networks/f5-common-python>`_
+
+- `Learn more about iControl REST <https://devcentral.f5.com/Wiki/Default.aspx?Page=HomePage&NS=iControlREST>`_
+
+
+Connecting to BIG-IP
+````````````````````
+
+Any BIG-IP user with administrative rights can use the F5 Modules for Ansible.
+
+To secure the user's password so it is not stored in plain text in your playbook or inventory file, you can use `Ansible Vault <https://docs.ansible.com/ansible/2.4/vault.html>`_.
+
+You do not need to exchange key pairs between the machine running Ansible and BIG-IP, because the F5 modules use the API, rather than SSH, to connect.
+
+In addition, your playbook must contain ``delegate_to: localhost``, so that the modules run locally on the machine running Ansible. Without this setting, the modules try to run on BIG-IP and fail, because the supporting Python libraries are not on BIG-IP.
+
+
+Common parameters
+`````````````````
+
+Every F5 module accepts the following parameters:
+
+server
+   Host name or IP address of BIG-IP.
+
+user
+   The user who can connect to BIG-IP. This user must have administrative privileges.
+
+password
+   Password for the user. Use Ansible Vault to encrypt the password, rather than storing it as plain text in your playbook or inventory file.
+
+validate_certs
+   Use to validate self-signed SSL certificates on personally-controlled sites.
+
+
+Common Tasks
+````````````
+
+Use the F5 Modules for Ansible to perform actions against BIG-IP.
+
+For reference information for each module, see `this list <http://docs.ansible.com/ansible/latest/list_of_network_modules.html#f5>`_.
+
+License BIG-IP
+++++++++++++++
+
+If you have a BIG-IP license, you can use Ansible to license BIG-IP. This example shows the full playbook.
+
+.. code-block:: yaml
+
+   ---
+
+   - name: License BIG-IP
+     hosts: f5-test
+     connection: local
+
+     vars:
+         bigip_license: "XXXXX-XXXXX-XXXXX-XXXXX-XXXXXXX"
+
+     tasks:
+         - name: License BIG-IP
+           bigip_license:
+               server: "{{ ansible_host }}"
+               user: "{{ bigip_username }}"
+               password: "{{ bigip_password }}"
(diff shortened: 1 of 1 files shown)


docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+About the API
+`````````````
+
+The primary API that's included with BIG-IP is called iControl REST. The Python library that interacts with iControl REST is called ``f5-sdk``.
+
+To work with the F5 Modules for Ansible, you should install ``f5-sdk``, as well as:
+
+- ``bigsuds``, the F5 SOAP API that was used for modules prior to Ansible 2.2
+- ``netaddr``, which helps with address manipulation
+- ``deepdiff``, to compare and find the differences between objects
```
caphrim007: I dont think this is a dependency any longer. Though it may still be listed as one in some docs. I'll remove it.

docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+
+- `Learn more about iControl REST <https://devcentral.f5.com/Wiki/Default.aspx?Page=HomePage&NS=iControlREST>`_
+
+
+Connecting to BIG-IP
+````````````````````
+
+Any BIG-IP user with administrative rights can use the F5 Modules for Ansible.
+
+To secure the user's password so it is not stored in plain text in your playbook or inventory file, you can use `Ansible Vault <https://docs.ansible.com/ansible/2.4/vault.html>`_.
```
caphrim007: Let's change this URL to

https://docs.ansible.com/ansible/latest/vault.html

So that we do not pin to a specific ansible release

docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+
+You do not need to exchange key pairs between the machine running Ansible and BIG-IP, because the F5 modules use the API, rather than SSH, to connect.
```
caphrim007: This is not entirely true. There is, at this time, one exception to this rule; the bigip_command module. Users of that module have the option of using either REST (default) or SSH to connect to the remote BIG-IP.

docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+
+In addition, your playbook must contain ``delegate_to: localhost``, so that the modules run locally on the machine running Ansible. Without this setting, the modules try to run on BIG-IP and fail, because the supporting Python libraries are not on BIG-IP.
```
caphrim007: The terminology here is a little off.

A Playbook, would contain `connection: local`.

Tasks of the playbook would contain `delegate_to: localhost`

The above are two ways of saying the same thing. The difference is that `connection: local` applies to all Tasks. Where-as `delegate_to: localhost` only applies to a single Task (and therefore must be re-specified for each Task you want to use it on).

docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+
+user
+   The user who can connect to BIG-IP. This user must have administrative privileges.
+
+password
+   Password for the user. Use Ansible Vault to encrypt the password, rather than storing it as plain text in your playbook or inventory file.
+
+validate_certs
+   Use to validate self-signed SSL certificates on personally-controlled sites.
+
```
caphrim007: Add `server_port` to this list

docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+For a full walkthrough of this example, `see this doc <http://clouddocs.f5.com/products/orchestration/ansible/devel/usage/playbook_tutorial.html>`_.
+
+.. code-block:: yaml
+
+     tasks:
+         - name: Create a pool
+           bigip_pool:
+               server: "{{ ansible_host }}"
+                  user: "{{ bigip_username }}"
```
caphrim007: `user` is indented too much here

docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+               password: "{{ bigip_password }}"
+               validate_certs: "{{ validate_certs }}"
+               lb_method: "ratio-member"
+               name: "web_pool"
+               slow_ramp_time: "120"
+           delegate_to: localhost
+
+         - name: Create node1
```
caphrim007: this can be removed. Nodes are created automatically when pool members are added.

docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+           bigip_node:
+               server: "{{ ansible_host }}"
+               user: "{{ bigip_username }}"
+               password: "{{ bigip_password }}"
+               validate_certs: "{{ validate_certs }}"
+               host: "10.10.10.10"
+               name: "node-1"
+           delegate_to: localhost
+
+         - name: Create node2
```
caphrim007: this can be removed. Nodes are created automatically when pool members are added.

docs/usage/guide_f5.rst
```diff
@@ -0,0 +1,316 @@
...
+.. code-block:: yaml
+
+         - name: run multiple commands on remote nodes
+           bigip_command:
+               server: "{{ ansible_host }}"
+               user: "{{ bigip_username }}"
+               password: "{{ bigip_password }}"
+               validate_certs: "{{ validate_certs }}"
+               commands:
+               - show sys version
```
caphrim007: list of commands should be indented under the `commands` key

docs/usage/guide_f5.rst
selhorn: Are you referring to just bigsuds, or are you referring to all three?

docs/usage/guide_f5.rst
selhorn: If they're created automatically, how do you know the node name when you create the pool? (Or is it not required? Should I drop name: node-1 and name: node-2 from the example?)

           bigip_pool_member:
               ...
               pool: web_pool
               port: 80
           delegate_to: localhost
           with_items:
               - host: 10.10.10.10
                 name: node-1
               - host: 10.10.10.20
                 name: node-2

docs/usage/guide_f5.rst
selhorn: OK, I've done that--I assume I should add it to all the examples, and that it's the server port for the BIG-IP config utility (8443 or 443 usually)-- ?

docs/usage/guide_f5.rst
selhorn: OK, thanks. Is this expected to change (like other modules will eventually have SSH)? If not, I will call out this one exception.

docs/usage/guide_f5.rst
caphrim007: @selhorn only deepdiff

docs/usage/guide_f5.rst
caphrim007: @selhorn it is a long-term goal that the other modules would also support SSH, but it is not currently on the roadmap

docs/usage/guide_f5.rst
caphrim007: @selhorn correct. Ansible will always default it to 443 though, at this time

caphrim007:
@selhorn you do not need quotes around those

docs/usage/guide_f5.rst
caphrim007: @selhorn you can get away with just specifying the host. you do not need the name unless you specifically want to name your nodes. By default their name is that of the ip address they are assigned.

For example,

```
    - name: Create some pool members
      bigip_pool_member:
        host: "{{ item.host }}"
        pool: my-pool
        port: "{{ item.port }}"
        rate_limit: 100
      loop:
        - host: 1.1.1.1
          port: 8443
        - host: 2.2.2.2
          port: 8080
```

PR #736, opened 2018-05-24
Updating support policy to refer to AskF5 KB and F5 policies.

rsalsa:
Updating SUPPORT file to reflect F5 policies for open source project as per AskF5 article [SOL80012344](https://support.f5.com/kb/en-us/solutions/public/k/80/sol80012344.html).

### DIFF
--- a/SUPPORT.rst
+++ b/SUPPORT.rst
@@ -1,24 +1,9 @@
 Support
 #######
 
-The code provided in this repository should be considered F5 contributed.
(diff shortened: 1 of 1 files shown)


caphrim007:
@rsalsa this removes most of the text that discusses the non-official-support channels. Being that this is an open source project, I think that non-official channel needs to be maintained. It should be spelled out that it is not the official support channel, but should not be removed.

The change here suggests that the ONLY means of support is via the official channels. That is entirely untrue.

PR #866, opened 2018-08-15
fixes #846

### LINKED ISSUE #846: bigip_profile_client_ssl other options
##### ISSUE TYPE

 - Feature Idea

##### COMPONENT NAME

bigip_profile_client_ssl 
##### ANSIBLE VERSION

```
ansible 2.6.2
```

##### PYTHON VERSION

```
Python 2.7.5
```

##### BIGIP VERSION

```
Viprion 2150:
  Product     BIG-IP
  Version     13.1.0.5
  Build       0.0.5
  Edition     Point Release 5
```

##### LIBRARY VERSIONS

```
bigsuds==1.0.6
f5-sdk==3.0.18
```

##### OS / ENVIRONMENT

N/A
##### SUMMARY

Add the following options to bigip_profile_client_ssl:

- allow_non_ssl: maps to allow-non-ssl in ltm config
- options: maps to options in ltm config.  Should allow an array or list of options that are pre-populated in the options list.
- secure_renegotiation: maps to secure-renegotiation in ltm config.
##### STEPS TO REPRODUCE

```
- name: Create a client SSL profile
  bigip_profile_client_ssl:
    state: present
    server: lb.mydomain.com
    user: admin
    password: secret
    name: my_profile
    cert_key_chain:
      - cert: bigip_ssl_cert1
        key: bigip_ssl_key1
        chain: bigip_ssl_cert1
    ciphers: "!SSLv3:!SSLv2:ECDHE+AES-GCM+SHA256:ECDHE-RSA-AES128-CBC-SHA"
    allow_non_ssl: enabled
    secure_renegotiation: request
    options:
      - dont-insert-empty-fragments
      - no-sslv2
      - no-sslv3
  delegate_to: localhost
```

##### ACTUAL RESULTS

```

```

wojtek0806:


library/modules/bigip_profile_client_ssl.py
```diff
@@ -68,6 +67,55 @@
...
+        connections. In this mode the system refuses connections to insecure servers,
+        and terminates existing SSL connections to insecure servers.
+    choices:
+      - require
+      - require-strict
+      - request
+  allow_non_ssl
+    description:
+      - Enables or disables acceptance of non-SSL connections.
+      - When creating a new profile, the setting is provided by the parent profile.
```
caphrim007: include mention of `type: bool` here as well as another key (like `description`)

library/modules/bigip_profile_client_ssl.py
```diff
@@ -330,6 +480,21 @@ def cert_key_chain(self):
...
+    def options(self):
+        if self.want.options is None:
+            return None
+        if self.have.options is None and self.want.options == '':
+            return None
+        if self.have.options is not None and self.want.options == '':
+            return []
+        if self.have.options is None:
+            return self.want.options
+        if set(self.want.options) == set(self.have.options):
```
caphrim007: this condition here can be removed. Its sufficient to check the `!=` condition only because at this point we've checked the "side cases" (what you have above) and all that is left is to check the not-equal case because, by default, if nothing is returned, then nothing will be reported.

test/integration/targets/bigip_profile_client_ssl/tasks/issue-00846.yaml
```diff
@@ -0,0 +1,149 @@
+---
+
+- name: Issue 00846 - Include issue variables
+  include_vars:
+    file: issue-00846.yaml
+
+- name: Issue 00846 - Include issue variables
```
caphrim007: duplicate include

wojtek0806:
@caphrim007 changes pushed

library/modules/bigip_profile_client_ssl.py
wojtek0806: not sure what you meant so i just added ":" after allow_non_ssl so it looks like this:
````
  allow_non_ssl:
    description:
      - Enables or disables acceptance of non-SSL connections.
      - When creating a new profile, the setting is provided by the parent profile.
    type: bool
    version_added: 2.7
````

PR #891, opened 2018-08-28
removed SDK from bigip_monitor_https, converted classes to reflect current standards

wojtek0806:
removed SDK, converted classes to reflect current standards


wojtek0806:
i noticed something i want to check do not merge yet

wojtek0806:
ok it seems what i wanted to change would not work (I wanted to remove those if statements from create and fold it into UsableChanges but this would affect updating not only creating, anyway im ok to merge now

wojtek0806:
ill add _set_default_creation_values() as it exists in tcp monitor module

PR #892, opened 2018-08-28
removed SDK from bigip_monitor_snmp_dca, converted classes to reflect current standards

wojtek0806:


wojtek0806:
ill add _set_default_creation_values() as it is in tcp monitor

PR #895, opened 2018-08-28
removed SDK from bigip_monitor_tcp_half_opemn, converted classes to reflect current standards

wojtek0806:


wojtek0806:
for some reason revert did not work, anyway this seems to be merged in devel so we can close

PR #920, opened 2018-09-10
SDK dependency removal bigip_pool_member module

wojtek0806:
removing SDK dependency from module
adding cleanup to provider tests


library/modules/bigip_pool_member.py
```diff
@@ -603,25 +597,48 @@ def present(self):
             return self.create()
 
     def exists(self):
+        tmp_uri = self.pool_exist()
```
caphrim007: I don't like the way this reads. Replace it with the normal method of building the URI and checking it like you did in the `pool_exists` method instead of using the pool_exists method to return a value.

PR #967, opened 2018-10-01
refactor of bigip_device_dns module

wojtek0806:
- reworking module to fit current class system
- removing redundant parameters
- removing SDK from module

I had to rework most of the items and removed stuff that was unnecessary or seemed unfinished. All tests pass.

### DIFF
--- a/library/modules/bigip_device_dns.py
+++ b/library/modules/bigip_device_dns.py
@@ -55,6 +55,7 @@ options:
 extends_documentation_fragment: f5
 author:
   - Tim Rupp (@caphrim007)
+  - Wojciech Wypior (@wojtek0806)
 '''
 
 EXAMPLES = r'''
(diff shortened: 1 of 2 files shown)


library/modules/bigip_device_dns.py
```diff
@@ -104,153 +105,252 @@
...
 
-    @property
-    def dhcp(self):
-        valid = ['enable', 'enabled']
-        return True if self._values['dhcp'] in valid else False
-
     @property
     def ip_version(self):
-        if self._values['ip_version'] in [6, '6', 'options inet6']:
+        if self._values['ip_version'] == 6:
```
caphrim007: This change here drops the normalizing of certain API params. I dont think this is intentional?

library/modules/bigip_device_dns.py
```diff
@@ -104,153 +105,252 @@
...
             return "options inet6"
-        elif self._values['ip_version'] in [4, '4', '']:
+        elif self._values['ip_version'] == 4:
```
caphrim007: same here. this was normalizing an api param

library/modules/bigip_device_dns.py
wojtek0806: i have never seen the other params, this is correct, the only thing bigip returns is options inet6, so that was redundant

library/modules/bigip_device_dns.py
wojtek0806: see my comment to first change

wojtek0806:
need to do one more thing here in regards to absenting parameters

PR #1100, opened 2018-12-04
fixes #880

### LINKED ISSUE #880: allow bigiq_regkey_license_assignment module to optionally assign next reg key from pool
- Feature Idea:

When licensing a device, the playbook must provide the registration key explicitly. It would be useful if there was an option to leave the the registration key empty, and allow the module to assign the next available key from the pool.

wojtek0806:


caphrim007:
@wojtek0806 think you need to rename the file. the file you have is called bigiq_license, but the docs say bigiq_regkey_license

wojtek0806:
@caphrim007 see teams

PR #1246, opened 2019-02-28
typo secure_renegotiation should be secure_renegotiation

lopf:


lopf:
included typo fix in https://github.com/F5Networks/f5-ansible/pull/1247 already

PR #1360, opened 2019-06-05
fix check mode in bigip_selfip on create

mhoertzsch:
selfip is created in check mode running ansible with -C option, since the wrong property is checked.


wojtek0806:
i have a fix already for this with some other changes so this is unnecessary, thank you for your submission tho!

PR #1402, opened 2019-07-04
Match all white space after prompt

atopuzov:
I encountered a timeout issue with Ansible `bigip_command` where it would get stuck because it would not detect the prompt. Oddly enough devices that were problematic were all `(cfg-sync Changes Pending)(Active)` state and they would send 2 additional characters, space and a carriage return.

Problematic device `(cfg-sync Changes Pending)(Active)(/Common)(tmos)#  ^M`
Hex for the last few chars: `29:23:20:20:0d`

Working device `(cfg-sync Changes Pending)(Standby)(/Common)(tmos)# `
Hex for the last few chars: `29:23:20`

From http://www.asciitable.com/
image

### DIFF
--- a/library/plugins/terminal/bigip.py
+++ b/library/plugins/terminal/bigip.py
@@ -28,9 +28,9 @@ from ansible.errors import AnsibleConnectionFailure
 class TerminalModule(TerminalBase):
 
     terminal_stdout_re = [
-        re.compile(br"[\r\n]?(?:\([^\)]+\)){,5}(?:>|#) ?$"),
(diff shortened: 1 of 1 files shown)


wojtek0806:
Hello, as we slowly move towards Ansible collections, we have stopped accepting external contributions, we encourage users to open up issues instead. As you never submitted PR here before you have never signed a CLA document required by external contributors, I could not accept this PR even as an exception. This is because the this document is no longer available as it has been removed along with our development documentation. What I will do is open an issue and link this PR to give you some credit of finding this.

PR #1424, opened 2019-07-23
Fixes #1364

### LINKED ISSUE #1364: collect user information with bigip_device_info
to help with getting user facts as per: https://github.com/F5Networks/f5-ansible/issues/1362 we need to add bigip user information to bigip_device_info module

nitinthewiz:
I've left out the encrypted password as it doesn't make sense to include that information in the response. Primary use of bigip_device_info -> users  is to allow customers to track the users that exist.

### DIFF
--- a/library/modules/bigip_device_info.py
+++ b/library/modules/bigip_device_info.py
@@ -90,6 +90,7 @@ options:
       - traffic-groups
       - trunks
       - udp-profiles
+      - users
       - vcmp-guests
       - virtual-addresses
       - virtual-servers
@@ -153,6 +154,7 @@ options:
       - "!traffic-groups"
       - "!trunks"
       - "!udp-profiles"
+      - "!users"
       - "!vcmp-guests"
       - "!virtual-addresses"
       - "!virtual-servers"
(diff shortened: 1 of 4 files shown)


library/modules/bigip_device_info.py
```diff
@@ -6088,6 +6090,47 @@
       type: bool
       sample: yes
   sample: hash/dictionary of values
+users:
```
wojtek0806: maybe also add there what shell the user can access, to be in line with what the bigip_user returns too:

https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_user.py#L200

test/integration/targets/bigip_device_info/tasks/issue-01364.yaml
```diff
@@ -0,0 +1,15 @@
+---
+
+- name: Issue 01364 - Select users facts
+  bigip_device_info:
+    include:
+      - users
+  register: result
+
+# - name: Issue 00082 - Assert Select interface facts
```
wojtek0806: incorrect checks, complete the test, i would suggest, tests like this: create a resource and then get it with facts and verify it is created

nitinthewiz: Fixed up the checks. Added the check as you suggested.

library/modules/bigip_device_info.py
nitinthewiz: Added. I was not sure how the shell works. Now, I'm using it as part of the checks!

nitinthewiz:
Now with 100% fewer whitespaces!

Issue #1431, opened 2019-07-25
bigip_profile_client_ssl not idempotent in BIG-IP 14.1


jolson490:
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME

bigip_profile_client_ssl

##### ANSIBLE VERSION

```
20:20:57 ansible-playbook 2.8.2
20:20:57   config file = /home/jenkins/workspace/cdpipeline/f5-management/sandbox/converge_ssl_profiles/ansible.cfg
20:20:57   configured module search path = [u'/home/jenkins/workspace/cdpipeline/f5-management/sandbox/converge_ssl_profiles/lib']
20:20:57   ansible python module location = /usr/local/lib/python2.7/dist-packages/ansible
20:20:57   executable location = /usr/local/bin/ansible-playbook
20:20:57   python version = 2.7.13 (default, Sep 26 2018, 18:42:22) [GCC 6.3.0 20170516]
```

##### PYTHON VERSION

```
Python 2.7.13
```

##### BIGIP VERSION

```
BIG-IP 14.1.0.3 Build 0.0.6 Point Release 3
```

##### ROLE VERSION

```

```

##### CONFIGURATION

n/a

##### OS / ENVIRONMENT

Linux

##### SUMMARY

After stepping up from BIG-IP version `14.0.0.1` to `14.1.0.3`, now when I run the following `ansible-playbook` command multiple times in a row, after the first time I would expect there to be no changes, but there's changes every time.

Snippet from my playbook yml file:
```
# Reference info: https://clouddocs.f5.com/products/orchestration/ansible/devel/modules/bigip_profile_client_ssl_module.html
- name: "Create profile {{ ssl_profile.name }} on {{ ec2_ip }}"
  bigip_profile_client_ssl:
    provider: "{{ f5_provider }}"
    name: "{{ ssl_profile.name }}"
    cert_key_chain: "{{ ssl_profile.cert_key_chain }}"
    allow_non_ssl: no
    options:
    - dont-insert-empty-fragments
    renegotiation: no
```

##### STEPS TO REPRODUCE

```
20:20:57 + ansible-playbook -i ec2.py f5-ssl-profiles.yml --limit tag_Cluster_sandbox -vvvv

20:20:57 host_list declined parsing /home/jenkins/workspace/cdpipeline/f5-management/sandbox/converge_ssl_profiles/ec2.py as it did not pass it's verify_file() method

[...]

PR #1438, opened 2019-07-31
Parse new `options` format from v14. Fix for #1431

dbanttari:


wojtek0806:
thank you for your interest, we no longer accept external submissions

wojtek0806:
Looks like we spoke too soon, before we start testing this, follow this document:

https://clouddocs.f5.com/products/orchestration/ansible/devel/development/cla-landing.html

dbanttari:
Document submitted~

jolson490:
note that this PR resolves https://github.com/F5Networks/f5-ansible/issues/1431

dbanttari:
Looks like there may be issues with existing integration tests~
```
TASK [bigip_profile_client_ssl : Create a client SSL profile] ***********************************************************
fatal: [bigip1]: FAILED! => {"changed": false, "msg": "01071b87:3: Profile /Common/foo cannot contain both ciphers and a cipher-group."}
```

wojtek0806:
this means this change introduced a regression, until this is resolved we will not merge this

dbanttari:
> this means this change introduced a regression, until this is resolved we will not merge this

This is _entirely_ unrelated to my change, and works fine against a v12.1 device.  What this "means" is that these tests have never been run against a v14.1 device.  If it doesn't take too much time i'll try to fix the v14 test regressions, starting with this one, but I'm not on your payroll so I have to time-box this.

dbanttari:
Found the source -- it's not a regression in the test, it's because we switched "clientssl" (on the v14 device i'm testing with) to use a cipherGroup instead of a list of ciphers, and that's causing problems with certain tests that are unrelated to my code change.  When I skip those tests it's fine.

wojtek0806:
just write the tests for your change i will get to this next week.

dbanttari:
Actually, it looks like the tests for issue #846 already catch this problem:
```
TASK [bigip_profile_client_ssl : Issue 00846 - Create a client SSL profile with options  - Idempotent check] ************************************************************************
changed: [bigip14]

TASK [bigip_profile_client_ssl : Issue 00846 - Assert Create a client SSL profile with options - Idempotent check] ******************************************************************
fatal: [bigip14]: FAILED! => {
    "assertion": "result is not changed",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```
and with my change in place:
```
TASK [bigip_profile_client_ssl : Issue 00846 - Create a client SSL profile with options  - Idempotent check] ************************************************************************
ok: [bigip14]

TASK [bigip_profile_client_ssl : Issue 00846 - Assert Create a client SSL profile with options - Idempotent check] ******************************************************************
ok: [bigip14] => {
    "changed": false,
    "msg": "All assertions passed"
}
```
I did run into an issue where some of the options that were being tested didn't exist on our v14 device, so I changed them to use different options.  Also added a few minor cleanup issues -- made the 00739 test use more-unique names, added a missing assertion to 00846, fixed a copy/paste-o in 00905, and used a more-likely-to-be-unique route domain number for 00532.

dbanttari:
for sake of completeness, all tests passed on a v12 device using the new code:
```
[massive SNIP]
TASK [bigip_profile_client_ssl : Issue 01250 - Cleanup] ***************************************************************************************************
changed: [bigip12]

PLAY RECAP ***************************************************************************************************
bigip12                    : ok=259  changed=94   unreachable=0    failed=0    skipped=0    rescued=0    ignored=3   

real	6m32.417s
```

test/integration/ansible.cfg
```diff
@@ -1,10 +1,10 @@
 [defaults]
 retry_files_enabled = False
 host_key_checking = False
-roles_path = /here/test/integration/targets/
-library = /here/library/modules
-module_utils = /here/library/module_utils
-lookup_plugins = /here/library/plugins/lookup
-action_plugins = /here/library/plugins/action
-terminal_plugins = /here/library/plugins/terminal
+roles_path = targets/
```
wojtek0806: can you revert this one, we use these ansible.cfg in our containers so the syntax should remain as it was, other than that this looks good

dbanttari: Done!

m-kratochvil:
Just to clarify - the use of "cipher-groups" is mandatory only when TLS 1.3 is enabled, which was added in v14.0. If the ssl profile option `no-tlsv1.3` is enabled, cipher-groups do not have to be used.

If the `bigip_profile_client_ssl` is run against v14.X device, WITHOUT the `no-tlsv1.3` option selected, execution fails with:
```fatal: [f511-02 -> localhost]: FAILED! => {"changed": false, "msg": "01b40001:3: A cipher group must be configured when TLS 1.3 is enabled (validation failed for profile /Common/foo_clientssl)."}```

Subsequently, if the `bigip_profile_client_ssl` is run against pre-v14 device, WITH the `no-tlsv1.3` option selected, module fails with:
```fatal: [f511-01 -> localhost]: FAILED! => {"changed": false, "msg": "\"options\", invalid value \"no-tlsv1.3\""}```

I would suggest adding a note to the "Comments" of the `options` parameter in the module page, specifying that the `no-tlsv1.3` option is valid only from v14.X up.

@wojtek0806

Issue #1381, opened 2019-06-14
bipip_node creation error "msg": "No JSON object could be decoded" when node name contains %


adamfitz:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

bigip_node

##### ANSIBLE VERSION

```
ansible 2.8.0
  config file = /home/adam/pool/ansible.cfg
  configured module search path = [u'/home/adam/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.5 (default, Apr  9 2019, 14:30:50) [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]

```

##### PYTHON VERSION

```
Python 2.7.5
```

##### BIGIP VERSION

```
Sys::Version
Main Package
  Product     BIG-IP
  Version     12.1.3.4
  Build       0.0.2
  Edition     Point Release 4
  Date        Tue Apr 10 14:26:02 PDT 2018

```

##### ROLE VERSION

```
Role: f5devcentral.f5ansible
        description: Ansible role that allows use of the F5 Ansible Modules
        active: True
        commit: 2a327bd27251dfcb2da3eff68827bcd675b616a4
        commit_message: Merging upstream changes - 2019-06-13T12:57:14Z
        commit_url: https://api.github.com/repos/f5devcentral/ansible-role-f5ansible/git/commits/2a327bd27251dfcb2da3eff68827bcd675b616a4
        company: F5 Networks
        created: 2018-10-04T16:14:10.510085Z
        dependencies: []
        download_count: 6970
        forks_count: 0
        galaxy_info:
                author: caphrim007
                company: F5 Networks
                galaxy_tags: ['f5', 'networking', 'bigip', 'bigiq']
                license: GPLv3
                min_ansible_version: 2.8
                platforms: [{'name': 'EL', 'versions': ['all']}, {'name': 'Fedora', 'versions': ['all']}, {'name': 'Ubuntu', 'versions': ['all']}, {'name': 'D
                role_name: f5ansible
        github_branch: master
        github_repo: ansible-role-f5ansible
        github_user: f5devcentral
        id: 30341
        imported: 2019-06-13T09:01:54.430041-04:00
        install_date: Fri Jun 14 05:40:35 2019
        installed_version: v2019.6.7
        is_valid: True
        issue_tracker_url:
[...]

PR #1467, opened 2019-08-21
integration test for issue #1381

nitinthewiz:
This confirms that creating a node with route domain works based on earlier fixes.


wojtek0806:
@crosbygw please review this today :)

crosbygw:
let's add a Idempotent check to make sure the added route domain is being returned as expected.

PR #1477, opened 2019-09-04
Added ansible_facts return variable with ansible_net_ prefix.

traittinen:
Added ansible_facts return variable on module.exit_json() to implement better support e.g. for facts caching in Ansible Tower and removing need to register variables. 

Also prepended all fact keys with ansible_net_ prefix. Old style device information is also returned for backward compatibility.

### DIFF
--- a/library/modules/bigip_device_info.py
+++ b/library/modules/bigip_device_info.py
@@ -16257,7 +16257,14 @@ def main():
     try:
         mm = ModuleManager(module=module)
         results = mm.exec_module()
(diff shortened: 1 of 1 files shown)


wojtek0806:
@traittinen we do not accept external submissions anymore, please file an issue that describes the problem and we will address this. Also any changes to the modules require functional tests attached.

wojtek0806:
@traittinen will credit you in our own PR that will have other fixes so your work won't go to waste

PR #1499, opened 2019-09-24
adding type to bigip_wait, including VCMP as an option

nitinthewiz:
First PR. Just want to know if I'm on the right track.

### DIFF
--- a/library/modules/bigip_wait.py
+++ b/library/modules/bigip_wait.py
@@ -23,6 +23,18 @@ description:
     of rebooting due to a configuration change.
 version_added: 2.5
 options:
+  type:
+    description:
+      - The type of the BIG-IP.
+      - Defaults to C(standard), the other choice is C(vcmp).
+      - The choice made defines what module or service Ansible will look for to establish
+        that the device has recovered, so please ensure the correct choice is
+        specified, specially when running this against VCMP.
+    type: str
+    default: standard
+    choices:
+      - standard
+      - vcmp
   timeout:
     description:
       - Maximum number of seconds to wait for.
@@ -210,8 +222,11 @@ class ModuleManager(object):
                     # of the waiting.
                     continue
 
-                if self._is_mprov_running_on_device():
-                    self._wait_for_module_provisioning()
+                if self.want.type == "standard":
+                    if self._is_mprov_running_on_device():
+                        self._wait_for_module_provisioning()
+                elif self.want.type == "vcmp":
+                    self._is_vcmpd_running_on_device()
                 break
             except Exception as ex:
                 if 'Failed to validate the SSL' in str(ex):
@@ -319,11 +334,39 @@ class ModuleManager(object):
             return True
         return False
 
+    def _is_vcmpd_running_on_device(self):
+        params = {
+            "command": "run",
+            "utilCmdArgs": '-c "ps aux | grep \'[v]cmpd\'"'
+        }
+        uri = "https://{0}:{1}/mgmt/tm/util/bash".format(
+            self.client.provider['server'],
+            self.client.provider['server_port']
+        )
+        resp = self.client.api.post(uri, json=params)
+        try:
+            response = resp.json()
+        except ValueError as ex:
+            raise F5ModuleError(str(ex))
+        if 'code' in response and response['code'] in [400, 403]:
+            if 'message' in response:
+                raise F5ModuleError(response['message'])
+            else:
+                raise F5ModuleError(resp.content)
+
+        if 'commandResult' in response:
+            return True
+        return False
+
 
 class ArgumentSpec(object):
     def __init__(self):
         self.supports_check_mode = True
         argument_spec = dict(
+            type=dict(
+                choices=['standard', 'vcmp'],
+                default='standard'
+            ),
             timeout=dict(default=7200, type='int'),
             delay=dict(default=0, type='int'),
             sleep=dict(default=1, type='int'),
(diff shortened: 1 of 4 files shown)


library/modules/bigip_wait.py
```diff
@@ -134,6 +141,12 @@ def timeout(self):
             return None
         return int(self._values['timeout'])
 
+    @property
```
wojtek0806: remove not needed, you have already specified defaults in the arg spec

library/modules/bigip_wait.py
```diff
@@ -319,11 +335,36 @@ def _is_mprov_running_on_device(self):
...
+        if 'commandResult' in response:
+            return True
+        return False
+
 
 class ArgumentSpec(object):
     def __init__(self):
         self.supports_check_mode = True
         argument_spec = dict(
+            type=dict(default='standard', type='str'),
```
wojtek0806: you are missing choices here, if you have a set of options just put them in choices and supplement documentation, do not let users guess what they supposed to enter, for example albeit with just one choice for now:

https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_message_routing_peer.py#L620

library/modules/bigip_wait.py
```diff
@@ -23,6 +23,13 @@
     of rebooting due to a configuration change.
 version_added: 2.5
 options:
+  type:
+    description:
+      - The type of the BIGIP defaults to standard, the other option being VCMP.
```
wojtek0806: mention the choices and what each choice is for, something like this:

https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_cgnat_lsn_pool.py#L48

wojtek0806:
please rebase this and do the changes so we can merge :) @nitinthewiz

library/modules/bigip_wait.py
nitinthewiz: Removed

library/modules/bigip_wait.py
nitinthewiz: Added the choices. Good recommendation. I'll make sure to use choices from here on out.

library/modules/bigip_wait.py
nitinthewiz: Updated docs.

library/modules/bigip_wait.py
```diff
@@ -23,6 +23,18 @@
     of rebooting due to a configuration change.
 version_added: 2.5
 options:
+  type:
+    description:
+      - The type of the BIGIP.
+      - Defaults to standard, the other choice being VCMP.
```
wojtek0806: should be C(standard) and C(vcmp), also do not capitalize parameters unless you really require them all in capital letters.

library/modules/bigip_wait.py
```diff
@@ -23,6 +23,18 @@
...
+      - The choice made defines what module/service Ansible will look for to establish
+      - that the device has recovered, so please ensure the correct choice is
+      - specified, specially when running this against VCMP.
```
wojtek0806: No need for hyphens, if this is a single passage you just need to use full stops and place the sentences one after the other, example here: https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_monitor_dns.py#L79

Also if you intend to keep `/` you need to wrap the entire string with double quotes otherwise YAML generators will go nuts.

library/modules/bigip_wait.py
nitinthewiz: Fixed. My explanation is a little more flowy. I hope that's fine.

library/modules/bigip_wait.py
nitinthewiz: Removed the /
Thanks for the link to the example. I was wondering about that, but took the safe option and put hyphens everywhere!

library/modules/bigip_wait.py
```diff
@@ -94,10 +106,10 @@
...
     from library.module_utils.network.f5.common import f5_argument_spec
 except ImportError:
-    from ansible_collections.f5networks.f5_modules.plugins.module_utils.bigip import F5RestClient
-    from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import F5ModuleError
-    from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import AnsibleF5Parameters
-    from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import f5_argument_spec
+    from ansible.module_utils.network.f5.bigip import F5RestClient
+    from ansible.module_utils.network.f5.common import F5ModuleError
+    from ansible.module_utils.network.f5.common import AnsibleF5Parameters
+    from ansible.module_utils.network.f5.common import f5_argument_spec
```
wojtek0806: change these import to what they were please

nitinthewiz: Interesting. I don't recall making these changes. Reverted.

library/modules/bigip_wait.py
```diff
@@ -319,11 +334,40 @@ def _is_mprov_running_on_device(self):
...
+
 
 class ArgumentSpec(object):
     def __init__(self):
         self.supports_check_mode = True
         argument_spec = dict(
+            type=dict(
+                choices=['standard', 'vcmp'],
+                default='standard',
+                type='str'
```
wojtek0806: no need to define str type in arg_spec remove :)

nitinthewiz: But isn't it good to define it as such?

wojtek0806: no it is defined explicitly in the documentation, arg_spec is not where it is placed

nitinthewiz: Done.

PR #1510, opened 2019-10-03
Updates to BIGIQ integration tests - FMFA-284

crosbygw:
- Added ability to use multiple bigips in bigiq integration tests
- Modified integration test vars to use bigiq_username to differentiate from bigip username
- added cleanup.py script which formats licenses into file used for integration tests.


test/integration/files/cleanup.py
```diff
@@ -0,0 +1,42 @@
...
+        for match in re.finditer(bigipLicenseRegex, line):
+            bigipLicense = (bigipLicense + "\n" + match.group())
+            print("Adding:" + match.group())
+    f_output = open(outputfile, "a+")
+    f_output.write(bigipLicense)
+    f_input = open(inputfile, "w")
+    f_input.write('')
+
+if __name__ == "__main__":
+    main(sys.argv[1:])
```
wojtek0806: nice touch! :) where is it run exactly?

.gitignore
```diff
@@ -26,8 +26,12 @@ docs/_modules/*_module.rst
 
 devtools/docker-compose.site.yaml
 
+# intergration test files
 test/integration/inventory/hosts
 test/integration/inventory/group_vars/all.yaml
+bigiq-regkey-licenses.txt
+bigiq-regkey-licenses-unformated.txt
+bigiq-utility-licenses.txt
```
wojtek0806: i think we have secrets addon for github that should prevent this? Can you check if you got it installed? I think you can find it in out gitlab,  for now this can remain

test/integration/files/cleanup.py
crosbygw: It can be ran anywhere as long as the appropriate input/output file paths are sent when running the script.

PR #1536, opened 2019-11-05
Add persistence enable-disable to wide_ip

nitinthewiz:
Adds missing persistence and related persist_cidr_ipv4, persist_cidr_ipv6, and persistence_ttl options to the wide_ip

### DIFF
--- a/library/modules/bigip_gtm_wide_ip.py
+++ b/library/modules/bigip_gtm_wide_ip.py
@@ -119,6 +119,31 @@ options:
         module.
     type: str
     version_added: 2.8
+  persistence:
+    description:
+      - When C(yes), ensures that when a local DNS makes repetitive requests on
+        behalf of a client, the system reconnects the client to the same resource
+        as previous requests.
+      - When C(no), ensures repetitive requests do not reconnect the client
+        to the same resource.
+    type: bool
+  persistence_ttl:
+    description:
+      - Specifies the time to maintain a connection between an local DNS and
+        a particular virtual server.
+    type: int
+  persist_cidr_ipv4:
+    description:
+      - Specifies a mask used to group IPv4 LDNS addresses. This feature
+        allows one persistence record to be shared by LDNS addresses
+        that match within this mask.
+    type: int
+  persist_cidr_ipv6:
+    description:
+      - Specifies a mask used to group IPv6 LDNS addresses. This feature
+        allows one persistence record to be shared by LDNS addresses
+        that match within this mask.
+    type: int
 notes:
   - Support for TMOS versions below v12.x has been deprecated for this module, and will be removed in Ansible 2.12.
 extends_documentation_fragment: f5networks.f5_modules.f5
@@ -186,6 +211,22 @@ EXAMPLES = r'''
       password: secret
       server: lb.mydomain.com
   delegate_to: localhost
+
+- name: Assign a pool with persistence to the Wide IP
+  bigip_gtm_wide_ip:
+    pool_lb_method: round-robin
+    name: my-wide-ip.example.com
+    pools:
+      - name: pool1
+        persistence: yes
+        persist_cidr_ipv4: 24
+        persist_cidr_ipv6: 120
+        persistence_ttl: 3500
+    provider:
+      user: admin
+      password: secret
+      server: lb.mydomain.com
+  delegate_to: localhost
 '''
 
 RETURN = r'''
@@ -209,6 +250,26 @@ aliases:
   returned: changed
   type: list
   sample: ['alias1.foo.com', '*.wildcard.domain']
+persistence:
+  description: Whether the pool connections will be persisted.
+  returned: changed
+  type: bool
+  sample: False
+persist_cidr_ipv4:
+  description: Specifies a mask used to group IPv4 LDNS addresses.
+  returned: changed
+  type: int
+  sample: 32
+persist_cidr_ipv6:
+  description: Specifies a mask used to group IPv6 LDNS addresses.
+  returned: changed
+  type: int
+  sample: 128
+persistence_ttl:
+  description: Specifies the persistence TTL between an local DNS and a particular virtual server.
+  returned: changed
+  type: int
+  sample: 3600
 '''
 
 from ansible.module_utils.basic import AnsibleModule
@@ -222,6 +283,7 @@ try:
     from library.module_utils.network.f5.common import AnsibleF5Parameters
     from library.module_utils.network.f5.common import fq_name
     from library.module_utils.network.f5.common import f5_argument_spec
+    from library.module_utils.network.f5.common import flatten_boolean
     from library.module_utils.network.f5.common import transform_name
     from library.module_utils.network.f5.common import is_valid_fqdn
     from library.module_utils.network.f5.icontrol import tmos_version
@@ -232,6 +294,7 @@ except ImportError:
     from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import AnsibleF5Parameters
     from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import fq_name
     from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import f5_argument_spec
+    from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import flatten_boolean
     from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import transform_name
     from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import is_valid_fqdn
     from ansible_collections.f5networks.f5_modules.plugins.module_utils.icontrol import tmos_version
@@ -243,6 +306,9 @@ class Parameters(AnsibleF5Parameters):
         'poolLbMode': 'pool_lb_method',
         'rules': 'irules',
         'lastResortPool': 'last_resort_pool',
+        'persistCidrIpv4': 'persist_cidr_ipv4',
+        'persistCidrIpv6': 'persist_cidr_ipv6',
+        'ttlPersistence': 'persistence_ttl',
     }
 
     updatables = [
@@ -254,6 +320,9 @@ class Parameters(AnsibleF5Parameters):
         'disabled',
         'aliases',
         'last_resort_pool',
+        'persist_cidr_ipv4',
+        'persist_cidr_ipv6',
+        'persistence_ttl',
     ]
 
     returnables = [
@@ -264,6 +333,10 @@ class Parameters(AnsibleF5Parameters):
         'irules',
         'aliases',
         'last_resort_pool',
+        'persistence',
+        'persist_cidr_ipv4',
+        'persist_cidr_ipv6',
+        'persistence_ttl',
     ]
 
     api_attributes = [
@@ -274,6 +347,10 @@ class Parameters(AnsibleF5Parameters):
         'rules',
         'aliases',
         'lastResortPool',
+        'persistence',
+        'ttlPersistence',
+        'persistCidrIpv4',
+        'persistCidrIpv6',
     ]
 
 
@@ -408,6 +485,47 @@ class ModuleParameters(Parameters):
             return ''
         self._values['aliases'].sort()
         return self._values['aliases']
+
+    @property
+    def persistence(self):
+        if self._values['persistence'] is None:
+            return None
+        result = flatten_boolean(self._values['persistence'])
+        if result is None:
+            return None
+        if result == 'yes':
+            return 'enabled'
+        return 'disabled'
+
+    @property
+    def persistence_ttl(self):
+        if self._values['persistence_ttl'] is None:
+            return None
+        if 0 <= self._values['persistence_ttl'] <= 4294967295:
+            return self._values['persistence_ttl']
+        raise F5ModuleError(
+            "Valid 'persistence_ttl' must be in range 0 - 4294967295."
+        )
+
+    @property
+    def persist_cidr_ipv4(self):
+        if self._values['persist_cidr_ipv4'] is None:
+            return None
+        if 0 <= self._values['persist_cidr_ipv4'] <= 4294967295:
+            return self._values['persist_cidr_ipv4']
+        raise F5ModuleError(
+            "Valid 'persist_cidr_ipv4' must be in range 0 - 4294967295."
+        )
+
+    @property
+    def persist_cidr_ipv6(self):
+        if self._values['persist_cidr_ipv6'] is None:
+            return None
+        if 0 <= self._values['persist_cidr_ipv6'] <= 4294967295:
+            return self._values['persist_cidr_ipv6']
+        raise F5ModuleError(
+            "Valid 'persist_cidr_ipv6' must be in range 0 - 4294967295."
+        )
 
 
 class Changes(Parameters):
@@ -922,6 +1040,10 @@ class ArgumentSpec(object):
                 type='list'
             ),
             last_resort_pool=dict(),
+            persistence=dict(type='bool'),
+            persistence_ttl=dict(type='int'),
+            persist_cidr_ipv4=dict(type='int'),
+            persist_cidr_ipv6=dict(type='int'),
         )
         self.argument_spec = {}
         self.argument_spec.update(f5_argument_spec)
--- a/test/integration/targets/bigip_gtm_wide_ip/tasks/environment/issue-persistence.yaml
+++ b/test/integration/targets/bigip_gtm_wide_ip/tasks/environment/issue-persistence.yaml
@@ -1,0 +1,108 @@
+---
+
+- name: Issue Persistence - Include issue variables
+  include_vars:
+    file: issue-persistence.yaml
+
+- name: Issue Persistence - Provision GTM
+  bigip_provision:
+    module: gtm
+  tags:
+    - module-provisioning
+
+- name: Issue Persistence - Create pool, type 'a'
+  bigip_gtm_pool:
+    preferred_lb_method: "{{ pool_lb_method }}"
+    name: "{{ pool_name }}"
+    state: present
+    type: a
+  register: result
+
+- name: Issue Persistence - Create wide IP with persistence
+  bigip_gtm_wide_ip:
+    lb_method: "{{ pool_lb_method }}"
+    state: present
+    type: a
+    name: "{{ wide_ip1 }}"
+    persistence: yes
+    persist_cidr_ipv4: 24
+    persist_cidr_ipv6: 120
+    pools:
+      - name: "{{ pool_name }}"
+  register: result
+
+- name: Issue Persistence - Assert Create wide IP with persistence
+  assert:
+    that:
+      - result is changed
+
+- name: Issue Persistence - Create wide IP with persistence - Idempotent check
+  bigip_gtm_wide_ip:
+    lb_method: "{{ pool_lb_method }}"
+    state: present
+    type: a
+    name: "{{ wide_ip1 }}"
+    persistence: yes
+    persist_cidr_ipv4: 24
+    persist_cidr_ipv6: 120
+    pools:
+      - name: "{{ pool_name }}"
+  register: result
+
+- name: Issue Persistence - Assert Create wide IP with persistence - Idempotent check
+  assert:
+    that:
+      - result is not changed
+
+
+- name: Issue Persistence - Create wide IP without persistence
+  bigip_gtm_wide_ip:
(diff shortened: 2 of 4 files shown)


wojtek0806:
once this PR is merged do full rebase and let the tests pass before we move on to this:

https://github.com/F5Networks/f5-ansible/pull/1540

nitinthewiz:
The build failed due to the following tests. This seems to be a part of the rebase, since my changes passed before the rebase. - 

ERROR: The 2 sanity test(s) listed below (out of 31) failed. See error output above for details.
action-plugin-docs
ignores

wojtek0806:
i need to see why do we get this error in shippable going to try a different branch

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+      - disabled
+      - enabled
+    default: disabled
+    version_added: 2.9
+  persistence_ttl:
+    description:
+      - Specifies the time to maintain a connection between an local DNS and
+        a particular virtual server. The default setting is 3600 seconds.
+    type: int
+    default: 3600
```
wojtek0806: no defaults

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+    version_added: 2.9
```
wojtek0806: incorrect version, 2.9 was already released

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+  persist_cidr_ipv4:
+    description:
+      - Specifies a mask used to group IPv4 LDNS addresses. This feature
+        allows one persistence record to be shared by LDNS addresses
+        that match within this mask. The default is 32.
+    type: int
+    default: 32
```
wojtek0806: no defaults

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+    version_added: 2.9
```
wojtek0806: incorrect version

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+      - Specifies the time to maintain a connection between an local DNS and
+        a particular virtual server. The default setting is 3600 seconds.
+    type: int
+    default: 3600
+    version_added: 2.9
+  persist_cidr_ipv4:
+    description:
+      - Specifies a mask used to group IPv4 LDNS addresses. This feature
+        allows one persistence record to be shared by LDNS addresses
+        that match within this mask. The default is 32.
```
wojtek0806: we do not inform what the default is by device as it may change version to version

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+    type: str
+    choices:
+      - disabled
+      - enabled
+    default: disabled
+    version_added: 2.9
+  persistence_ttl:
+    description:
+      - Specifies the time to maintain a connection between an local DNS and
+        a particular virtual server. The default setting is 3600 seconds.
```
wojtek0806: we do not inform what the default is by device as it may change version to version

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+  persistence:
+    description:
+      - When C(enabled), ensures that when a local DNS makes repetitive requests on
+        behalf of a client, the system reconnects the client to the same resource
+        as previous requests.
+      - When C(disabled), ensures repetitive requests do not reconnect the client
+        to the same resource.
+    type: str
+    choices:
+      - disabled
```
wojtek0806: this should be boolean, change it to boolean, look at other modules how to do this

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -209,6 +261,11 @@
   returned: changed
   type: list
   sample: ['alias1.foo.com', '*.wildcard.domain']
+persistence:
+  description: Whether the pool connections will be persisted.
+  returned: changed
+  type: str
```
wojtek0806: change to boolean

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+    type: int
+    default: 32
+    version_added: 2.9
+  persist_cidr_ipv6:
+    description:
+      - Specifies a mask used to group IPv6 LDNS addresses. This feature
+        allows one persistence record to be shared by LDNS addresses
+        that match within this mask. The default is 128.
+    type: int
+    default: 128
```
wojtek0806: no defaults

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+        allows one persistence record to be shared by LDNS addresses
+        that match within this mask. The default is 32.
+    type: int
+    default: 32
+    version_added: 2.9
+  persist_cidr_ipv6:
+    description:
+      - Specifies a mask used to group IPv6 LDNS addresses. This feature
+        allows one persistence record to be shared by LDNS addresses
+        that match within this mask. The default is 128.
```
wojtek0806: same as above

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,42 @@
...
+    type: int
+    default: 128
+    version_added: 2.9
```
wojtek0806: wrong version

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -922,6 +1017,22 @@ def __init__(self):
                 type='list'
             ),
             last_resort_pool=dict(),
+            persistence=dict(
+                default='disabled',
+                choices=['enabled', 'disabled']
+            ),
+            persistence_ttl=dict(
+                type='int',
+                default=3600
```
wojtek0806: remove defaults

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -922,6 +1017,22 @@ def __init__(self):
                 type='list'
             ),
             last_resort_pool=dict(),
+            persistence=dict(
+                default='disabled',
```
wojtek0806: no defaults

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -922,6 +1017,22 @@ def __init__(self):
...
+                choices=['enabled', 'disabled']
```
wojtek0806: should be boolean

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -922,6 +1017,22 @@ def __init__(self):
...
+                type='int',
+                default=3600
+            ),
+            persist_cidr_ipv4=dict(
+                type='int',
+                default=32
+            ),
+            persist_cidr_ipv6=dict(
+                type='int',
+                default=128
```
wojtek0806: remove

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -922,6 +1017,22 @@ def __init__(self):
...
+                default='disabled',
+                choices=['enabled', 'disabled']
+            ),
+            persistence_ttl=dict(
+                type='int',
+                default=3600
+            ),
+            persist_cidr_ipv4=dict(
+                type='int',
+                default=32
```
wojtek0806: remove

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +480,30 @@ def aliases(self):
         self._values['aliases'].sort()
         return self._values['aliases']
 
+    @property
+    def persistence(self):
+        if self._values['persistence'] is None:
+            return None
+        return str(self._values['persistence'])
```
wojtek0806: change

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +480,30 @@ def aliases(self):
...
+
+    @property
+    def persistence_ttl(self):
+        if self._values['persistence_ttl'] is None:
+            return None
+        return self._values['persistence_ttl']
+
+    @property
+    def persist_cidr_ipv4(self):
+        if self._values['persist_cidr_ipv4'] is None:
```
wojtek0806: why is this here? only use property when you are actually checking or doing value transformation. Also if you have int values usually its best to check if the number is valid, check the device documentation for valid range, as far as implementation goes you need  to look at other modules for established patterns use patterns

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +480,30 @@ def aliases(self):
...
+            return None
+        return self._values['persist_cidr_ipv4']
+
+    @property
+    def persist_cidr_ipv6(self):
```
wojtek0806: why is this here? only use property when you are actually checking or doing value transformation. Also if you have int values usually its best to check if the number is valid, check the device documentation for valid range, as far as implementation goes you need to look at other modules for established patterns use patterns

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +480,30 @@ def aliases(self):
...
+    @property
+    def persistence(self):
+        if self._values['persistence'] is None:
+            return None
+        return str(self._values['persistence'])
+
+    @property
+    def persistence_ttl(self):
+        if self._values['persistence_ttl'] is None:
+            return None
```
wojtek0806: why is this here? only use property when you are actually checking or doing value transformation. Also if you have int values usually its best to check if the number is valid, check the device documentation for valid range, as far as implementation goes you need to look at other modules for established patterns use patterns

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -209,6 +261,11 @@
   returned: changed
   type: list
   sample: ['alias1.foo.com', '*.wildcard.domain']
+persistence:
+  description: Whether the pool connections will be persisted.
+  returned: changed
+  type: str
+  sample: disabled
 '''
```
wojtek0806: where are the rest of returned values from returnables, document every value the module returns after it runs

wojtek0806:
Also no need to have provider/env type tests, just do the env type tests, some older modules had it due to some issues with arg.spec which has been removed. Also not having these tests means less work for us once we start tidying up those tests.

nitinthewiz:
> Also no need to have provider/env type tests, just do the env type tests, some older modules had it due to some issues with arg.spec which has been removed. Also not having these tests means less work for us once we start tidying up those tests.

I took a call because there was no documentation on which type of test is older. I'll just put my test outside of these.

wojtek0806:
all you need to do is look at the latest PRs all of them do not use this distinction

nitinthewiz:
I checked and the file in provider was there by mistake. The actual test is only under env.

library/modules/bigip_gtm_wide_ip.py
nitinthewiz: There are other scenarios, many of them, where the default has been specified. Removing regardless.

library/modules/bigip_gtm_wide_ip.py
nitinthewiz: changed. I was following the module's current pattern.

library/modules/bigip_gtm_wide_ip.py
nitinthewiz: added checking.

library/modules/bigip_gtm_wide_ip.py
nitinthewiz: added checking.

library/modules/bigip_gtm_wide_ip.py
nitinthewiz: added checking.

nitinthewiz:
Ok, shippable failed as expected and Travis passed.

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +484,50 @@ def aliases(self):
         self._values['aliases'].sort()
         return self._values['aliases']
 
+    @property
+    def persistence(self):
```
wojtek0806: This is incorrect, this is how you should do the boolean:

https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_profile_http.py#L1035

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -453,6 +572,15 @@ def last_resort_pool(self):
             return 'none'
         return self._values['last_resort_pool'].split(' ')[1]
 
+    @property
+    def persistence(self):
+        if self._values['persistence'] == 'enabled':
```
wojtek0806: why is this here? if API accepts enabled/disabled then this is not needed

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -119,6 +119,31 @@
         module.
     type: str
     version_added: 2.8
+  persistence:
+    description:
+      - When C(enabled), ensures that when a local DNS makes repetitive requests on
```
wojtek0806: booleans always flatten to yes/no not enabled/disabled

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -922,6 +1050,10 @@ def __init__(self):
                 type='list'
             ),
             last_resort_pool=dict(),
+            persistence=dict(type='bool', default=False),
```
wojtek0806: no defaults

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +484,50 @@ def aliases(self):
...
+            "Valid 'persist_cidr_ipv4' must be in range 0 - 4294967295."
+        )
+
+    @property
+    def persist_cidr_ipv6(self):
+        if not self._values['persistence']:
+            return 128
+        if self._values['persist_cidr_ipv6'] is None:
+            return None
+        if 0 <= int(self._values['persist_cidr_ipv6']) <= 4294967295:
```
wojtek0806: if this is an int why do you cast the value to an int?

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +484,50 @@ def aliases(self):
...
+        if 0 <= int(self._values['persist_cidr_ipv4']) <= 4294967295:
+            return int(self._values['persist_cidr_ipv4'])
+        raise F5ModuleError(
+            "Valid 'persist_cidr_ipv4' must be in range 0 - 4294967295."
+        )
+
+    @property
+    def persist_cidr_ipv6(self):
+        if not self._values['persistence']:
+            return 128
```
wojtek0806: why do you return values if they are not declared by the user? remove this

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +484,50 @@ def aliases(self):
...
+            "Valid 'persistence_ttl' must be in range 0 - 4294967295."
+        )
+
+    @property
+    def persist_cidr_ipv4(self):
+        if not self._values['persistence']:
+            return 32
+        if self._values['persist_cidr_ipv4'] is None:
+            return None
+        if 0 <= int(self._values['persist_cidr_ipv4']) <= 4294967295:
```
wojtek0806: if this is an int why do you cast the value to an int?

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +484,50 @@ def aliases(self):
...
+        if 0 <= int(self._values['persistence_ttl']) <= 4294967295:
+            return int(self._values['persistence_ttl'])
+        raise F5ModuleError(
+            "Valid 'persistence_ttl' must be in range 0 - 4294967295."
+        )
+
+    @property
+    def persist_cidr_ipv4(self):
+        if not self._values['persistence']:
+            return 32
```
wojtek0806: why do you return values if they are not declared by the user? remove this

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +484,50 @@ def aliases(self):
...
+    def persistence(self):
+        if self._values['persistence'] is None:
+            return None
+        if self._values['persistence']:
+            return 'enabled'
+        return 'disabled'
+
+    @property
+    def persistence_ttl(self):
+        if not self._values['persistence']:
```
wojtek0806: why do you return values if they are not declared by the user? remove this

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -409,6 +484,50 @@ def aliases(self):
...
+            return 3600
+        if self._values['persistence_ttl'] is None:
+            return None
+        if 0 <= int(self._values['persistence_ttl']) <= 4294967295:
```
wojtek0806: if this is an int why do you cast the value to an int?

library/modules/bigip_gtm_wide_ip.py
```diff
@@ -186,6 +211,22 @@
...
       server: lb.mydomain.com
   delegate_to: localhost
+
+- name: Assign a pool with persistence to the Wide IP
+  bigip_gtm_wide_ip:
+    pool_lb_method: round-robin
+    name: my-wide-ip.example.com
+    pools:
+      - name: pool1
+        persistence: True
```
wojtek0806: change

test/integration/targets/bigip_gtm_wide_ip/tasks/environment/issue-persistence.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+- name: Issue Persistence - Create wide IP without persistence
+  bigip_gtm_wide_ip:
+    lb_method: "{{ pool_lb_method }}"
+    state: present
+    type: a
+    name: "{{ wide_ip2 }}"
+    pools:
+      - name: "{{ pool_name }}"
+  when: f.system_info.product_version  >= "12.0.0"
```
wojtek0806: this kind of check needs to be done before the test is run

test/integration/targets/bigip_gtm_wide_ip/tasks/environment/issue-persistence.yaml
```diff
@@ -0,0 +1,124 @@
...
+- name: Issue Persistence - Provision GTM
+  bigip_provision:
+    module: gtm
+  tags:
+    - module-provisioning
+
+- name: Issue Persistence - Collect BIG-IP facts
+  bigip_device_facts:
+    include: system-info
+  register: f
```
wojtek0806: move it to main.yml and apply the version check to the entire task import, also gtm support for v11 is deprecated so it is not really needed

library/modules/bigip_gtm_wide_ip.py
nitinthewiz: Oh that's an interesting function for tackling boolean! I'll use it!

test/integration/targets/bigip_gtm_wide_ip/tasks/environment/issue-persistence.yaml
nitinthewiz: I'm just following one of the recent standards here - https://github.com/F5Networks/f5-ansible/blob/ba8e3002abe6d3075a611c7dc0427bb77ccb2281/test/integration/targets/bigip_network_globals/tasks/main.yaml

test/integration/targets/bigip_gtm_wide_ip/tasks/environment/issue-persistence.yaml
nitinthewiz: removed.

test/integration/targets/bigip_gtm_wide_ip/tasks/environment/issue-persistence.yaml
wojtek0806: this would make sense if you had a supported version in mind and some tasks were supposed to run on both versions, since you want all to run on v12 and beyond then it is not needed, also  v11 is deprecated so we can get rid of this :)

nitinthewiz:
I've deleted the product version check, as you recommended.

PR #1538, opened 2019-11-05
Fix for missing receive disable parameter in TCP monitor

nitinthewiz:


wojtek0806:
once this PR is merged do full rebase and let the tests pass before we move on to this:

https://github.com/F5Networks/f5-ansible/pull/1540

nitinthewiz:
I've added a test that removes the string and the result is accurate. So the code seems to handle that scenario on its own, based on the product's own process.

nitinthewiz:
Same - travis has passed and shippable failed.

PR #1552, opened 2019-11-11
Fixes #1551

### LINKED ISSUE #1551: Add mask and mirror options in bigip_profile_persistence_src_addr
##### ISSUE TYPE

 - Feature Idea

##### COMPONENT NAME

bigip_profile_persistence_src_addr

##### ANSIBLE VERSION

```
ansible 2.10.0.dev0
  config file = /root/nitin_ansible.cfg
  configured module search path = ['/root/f5-ansible/library/modules']
  ansible python module location = /usr/local/lib/python3.6/dist-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.6.8 (default, Oct  7 2019, 12:59:55) [GCC 8.3.0]
```

##### PYTHON VERSION

```
Python 3.6.8
```

##### BIGIP VERSION

```

Sys::Version
Main Package
  Product     BIG-IP
  Version     14.1.2.2
  Build       0.0.4
  Edition     Point Release 2
  Date        Tue Nov  5 10:09:50 PST 2019
```

##### SUMMARY

bigip_profile_persistence_src_addr should include the mask and mirror options. Mask should automatically convert IPv4 and IPv6 masks to the correct values.

##### STEPS TO REPRODUCE

```

```

##### ACTUAL RESULTS

```

```

nitinthewiz:
The webGUi allows the user to input a prefix length number (0-32 for IPv4 and 0-128 for IPv6) but it seems that the API expects it to be a proper mask, because there is no additional field for IP version in the API.

Thus, I've made is so that the mask will be a full mask of type '255.255.255.255' or 'ffff::'

### DIFF
--- a/library/modules/bigip_profile_persistence_src_addr.py
+++ b/library/modules/bigip_profile_persistence_src_addr.py
@@ -52,6 +52,19 @@ options:
       - When creating a new profile, if this parameter is not specified, the
         default is provided by the parent profile.
     type: bool
+  mirror:
+    description:
+      - When C(yes), specifies that if the active unit goes into the standby mode, the system
+        mirrors any persistence records to its peer.
+      - When creating a new profile, if this parameter is not specified, the
+        default is provided by the parent profile.
+    type: bool
+  mask:
+    description:
+      - Specifies a value that the system applies as the prefix length.
+      - When creating a new profile, if this parameter is not specified, the
+        default is provided by the parent profile.
+    type: str
   hash_algorithm:
     description:
       - Specifies the algorithm the system uses for hash persistence load balancing. The hash
@@ -107,6 +120,8 @@ EXAMPLES = r'''
     hash_algorithm: carp
     match_across_services: yes
     match_across_virtuals: yes
+    mirror: yes
+    mask: 255.255.255.255
     provider:
       password: secret
       server: lb.mydomain.com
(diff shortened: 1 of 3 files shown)


nitinthewiz:
@wojtek0806 in this module, everything is done is class Parameters, whereas we tend to do things in class ModuleParameters now. So I've done it in the new way and the tests pass fine.

Just curious what you think the better way to do this is.

library/modules/bigip_profile_persistence_src_addr.py
```diff
@@ -232,6 +255,32 @@ def parent(self):
...
+            return None
+        result = flatten_boolean(self._values['mirror'])
+        if result is None:
+            return None
+        if result == 'yes':
+            return 'enabled'
+        return 'disabled'
+
+    @property
+    def mask(self):
```
wojtek0806: i would suggest finding a way to accept the CIDR and full mask format as it is better not to limit users options. Look here for some inspiration: https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_apm_acl.py#L573

When you accept CIDR format just use the ipaddress lib functions to expand it to full format and send that to the API

nitinthewiz: The problem is that of specifying whether the inputed value is IPv4 or IPv6. If it is something like '16' then it could be either. 

In that case, I'm forced to ask the user to tell me which version they're using. Also would need us to return the same.

Instead, this method is directly in line with the BIG-IP API itself.

I'll try to allow for both prefix and CIDR, but I don't see a lot of value in that.

wojtek0806: nah ok you convinced me, do something like this instead:

https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_virtual_address.py#L388

nitinthewiz: done!

PR #1571, opened 2019-11-18
Fixes #1399

### LINKED ISSUE #1399: Enable no port to TACACS servers and  "accounting send-to-all-servers"
##### ISSUE TYPE

 - Feature Idea

##### COMPONENT NAME

bigip_device_auth

##### ANSIBLE VERSION

```
ansible 2.8.0
```

##### PYTHON VERSION

```
Python 2.7.15rc1
```

##### BIGIP VERSION

```
BIG-IP 12.1.3 Build 0.0.378 Final
```

##### ROLE VERSION

```
 f5devcentral.f5ansible (v2019.5.10)
```

##### CONFIGURATION

```
   [defaults]
   host_key_checking             = False
   ssh_args                      = -C -o ControlMaster=auto -o ControlPersist=60s 
```

##### OS / ENVIRONMENT

```
Distributor ID: Ubuntu
Description:    Ubuntu 18.04 LTS
Release:        18.04
```

##### SUMMARY

Consider adding "accounting send-to-all-servers" to the TACACS configuration and not specify port of the tacacs server. 

When I use bigip_device_auth with no server port, by default it sets 49163, but the authentication doesn't work neither if I precise de tacacs default port 49 instead. So, is it possible to provide servers without the default port as the steps below? And also  "accounting send-to-all-servers" ?

##### STEPS TO REPRODUCE

via command line: 
```

modify auth source type tacacs
create auth tacacs system-auth   protocol ip secret mYt0pS3cr3T! server add {10.10.10.10 10.10.10.20} service  ppp accounting send-to-all-servers
modify auth remote-user {   default-role admin remote-console-access tmsh }

```

or via Ansible: 

```
    - name: Setting TACACS+ as the source Authentication base
      bigip_device_auth:
         type: tacacs
         servers: ["{{ TACACS['auth_servers'] }}"
         secret: "{{ TACACS['auth_secret'] }}"
         use_for_auth: yes
         protocol_name: ip
         service_name: ppp
         state: present
      delegate_to: localhost    
```

##### EXPECTED RESULTS

list auth tacacs 
>
```
auth tacacs system-auth {
    accounting send-to-all-servers
    protocol ip
    secret ******************
    servers { 10.10.10.10 10.10.10.20}
    service ppp
}
```

##### ACTUAL RESULTS

```
 list auth tacacs
auth tacacs system-auth {
    protocol ip
    secret **************==
    servers { 10.10.10.10:49163 10.10.10.20:49163 }
    service ppp
}

```

nitinthewiz:


library/modules/bigip_device_auth.py
```diff
@@ -119,6 +119,19 @@
     choices:
       - use-first-server
       - use-all-servers
+  accounting:
```
crosbygw: should include version_added:

library/modules/bigip_device_auth.py
```diff
@@ -119,6 +119,19 @@
     choices:
```
crosbygw: Noticed port sub-option on line 55 says default: 49163 - is this still true?

test/integration/targets/bigip_device_auth/tasks/environment/issue-01399.yaml
```diff
@@ -0,0 +1,135 @@
...
+    type: local
+    use_for_auth: yes
+  register: result
+
+- name: Assert Check if local is set
+  assert:
+    that:
+      - result is not changed
+      - result is success
+
```
crosbygw: lets add a check to verify absence of server port works.

library/modules/bigip_device_auth.py
```diff
@@ -198,6 +212,11 @@
   returned: changed
   type: str
   sample: use-all-servers
+accounting:
+  description: Which servers to send information to when using TACACS.
+  returned: changed
+  type: str
+  sample: send-to-all-servers
 service_name:
```
crosbygw: please add yourself as author.

library/modules/bigip_device_auth.py
crosbygw: let's add unit test for new parameter.

nitinthewiz: I'm staying away from version added because, what's the answer? Collections 1.0?

library/modules/bigip_device_auth.py
nitinthewiz: Yeah I need to fix that!

test/integration/targets/bigip_device_auth/tasks/environment/issue-01399.yaml
```diff
@@ -0,0 +1,135 @@
+---
+
+- import_tasks: setup.yaml
```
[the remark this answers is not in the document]
nitinthewiz: tacacs actually ends with a cleanup that removes the setup steps. That's why I re-imported it.

test/integration/targets/bigip_device_auth/tasks/environment/issue-01399.yaml
nitinthewiz: There IS a check for this.
I've added all three possible types of server configs to the initial config - 
```
      - 10.10.10.10
      - address: 10.10.10.11
        port: 1234
      - 10.10.10.12:1245
```
and it passes perfectly.

library/modules/bigip_device_auth.py
nitinthewiz: Added

library/modules/bigip_device_auth.py
nitinthewiz: Done.

test/integration/targets/bigip_device_auth/tasks/environment/issue-01399.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - 10.10.10.12:1245
+    service_name: ppp
+    use_for_auth: yes
+    update_secret: on_create
+  register: result
+
+- name: Assert Set sys auth to TACACS+
+  assert:
+    that:
+      - result is changed
```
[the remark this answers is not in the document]
nitinthewiz: done!

test/integration/targets/bigip_device_auth/tasks/environment/issue-01399.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - result is not changed
+      - result is success
+
+- name: Change accounting, send-to-first-server
+  bigip_device_auth:
+    type: tacacs
+    accounting: send-to-first-server
+  register: result
+
+- name: Assert Change accounting, send-to-first-server
```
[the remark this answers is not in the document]
nitinthewiz: done

test/integration/targets/bigip_device_auth/tasks/environment/issue-01399.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - result is not changed
+      - result is success
+
+- name: Change accounting, send-to-all-servers
+  bigip_device_auth:
+    type: tacacs
+    accounting: send-to-all-servers
+  register: result
+
+- name: Assert Change accounting, send-to-all-servers
```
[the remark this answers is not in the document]
nitinthewiz: done

library/modules/bigip_device_auth.py
nitinthewiz: done

library/modules/bigip_device_auth.py
wojtek0806: add the option as requested, the answer to a format is in the code and in our stubber templates:

https://github.com/F5Networks/f5-ansible/blob/devel/devtools/stubs/library_module.py#L21
https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_cgnat_lsn_pool.py#L21
On a side note I will need to update the template to have 'f5_modules 1.1.0'

library/modules/bigip_device_auth.py
```diff
@@ -320,11 +342,15 @@ def servers(self):
                         "An 'address' field must be provided when specifying separate fields to the 'servers' parameter."
                     )
                 address = server.get('address')
-                port = server.get('port', 49163)
```
wojtek0806: while i understand why you did this (i.e no defaults), changing the 'default' now will cause issues for users that were relying on this in their plays. avoid doing such changes without adding deprecation warnings

wojtek0806: if this was needed cause the module was breaking then this kind of user breaking change would be acceptable

library/modules/bigip_device_auth.py
```diff
@@ -320,11 +342,15 @@ def servers(self):
...
+                if 'port' in server:
+                    port = server.get('port')
```
wojtek0806: this is unnecessary just define get() method to return None if there is no key e.g. 

`port = server.get('port', None)`

library/modules/bigip_device_auth.py
```diff
@@ -320,11 +342,15 @@ def servers(self):
...
             elif isinstance(server, string_types):
                 address = server
-                port = 49163
-            result.append('{0}:{1}'.format(address, port))
+                port = None
```
wojtek0806: see my comments above

library/modules/bigip_device_auth.py
```diff
@@ -52,7 +52,6 @@
       port:
         description:
           - The port of the server.
-        default: 49163
```
wojtek0806: i know i said no defaults but we just don't remove stuff without deprecation

wojtek0806:
change and rebase

library/modules/bigip_device_auth.py
nitinthewiz: it actually doesn't break user experience because the user doesn't want the port to be specified, and BIG-IP documentation does't recommend it. So the user will end up in a positive situation according to my read of the issue.

library/modules/bigip_device_auth.py
nitinthewiz: Hmmm sure.

library/modules/bigip_device_auth.py
nitinthewiz: done.

library/modules/bigip_device_auth.py
nitinthewiz: So what's the rule of thumb on how to deprecate a default?

wojtek0806: you would announce deprecations with this method:
https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_gtm_wide_ip.py#L743

There is no set recipe as each module does it differently but the idea is for the Ansible to announce deprecation at the end of module run. I keep the code relatively tech debit free so i removed some good examples already on how to do this.

PR #1637, opened 2020-02-04
Update bigip_policy_rule.py

BapRx:
Updated action "reset" with event "request"
Added action "disable" with target "server_ssl"


wojtek0806:
Hello, thank you for your work, as we do not accept external PR submissions we will have to reject this PR.

PR #1640, opened 2020-02-05
Fixes #1521

### LINKED ISSUE #1521: bigip_policy_rule: Add support for "forward node"
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
bigip_policy_rule

##### ANSIBLE VERSION
```
ansible 2.10.0.dev0
  config file = None
  configured module search path = 
  ansible python module location = 
  executable location = ansible-dev/bin/ansible
  python version = 2.7.16 (default, Aug 24 2019, 18:37:03) [GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.32.4) (-macos10.15-objc-s
```

##### PYTHON VERSION
```
see above
```

##### SUMMARY
The module doesn't support "forward node" as action. In our environment we're using forward to nodes in specific route domains. Therefore the route domain needs to be supported as well, either as part of the IP address or as a separate argument.

nitinthewiz:


library/modules/bigip_policy_rule.py
```diff
@@ -63,6 +63,11 @@
           - Virtual Server that you want to forward traffic to.
           - This parameter is only valid with the C(forward) type.
         type: str
+      node:
+        description:
+          - Node that you want to forward traffic to.
+          - This parameter is only valid with the C(forward) type.
+        type: str
```
wojtek0806: version added missing

nitinthewiz: Fixed! Also added my name to authors.

Issue #1483, opened 2019-09-11
Feature request: SSL Cipher groups


VeselaHouba:
Hi,
Do you think about implementing SSL Cipher groups assignment into client ssl profile? 
https://github.com/F5Networks/f5-ansible/blob/devel/library/modules/bigip_profile_client_ssl.py

And possibly also definition of the groups via ansible.

I know I should create a pull request with the code, but with my programming skills it would probably end-up with F5 box burning down in flames ;)

//JM

wojtek0806:
the fundamental question is are both of these achievable via REST api?

wojtek0806:
FMFA-410

PR #1641, opened 2020-02-06
Added support for cipher group to bigip_profile_client_ssl

apatik:
Addresses issue #1483


wojtek0806:
thank you for your interest, we do not accept external PR submissions, coincidently i have solution to this issue already but it is not pushed to PR yet, so i will merge mine in the next few days.

PR #1662, opened 2020-02-27
Include aliases to bigip_monitor_tcp module

andreykashcheev:
### Summary 

This change is intended to add aliases to ip and port fields to prevent confusions since these two fields are titled differently in BIG IP Management UI. Also, updated fields metadata to inform end-users about mis-match between Ansible Module parameters and BIG IP Management UI 

### Changes

 - Add aliases to ip and port module parameters
 - Updated description for ip and port module parameters

### Additional Information


library/modules/bigip_monitor_tcp.py
```diff
@@ -63,7 +66,10 @@
         '*'. Note that if specifying an IP address, a value between 1 and 65535
         must be specified
       - This argument is not supported for TCP Echo types.
+      - In BIG IP Management UI, this field is titled as 'Alias Service Port'
```
wojtek0806: missing fullstop

library/modules/bigip_monitor_tcp.py
```diff
@@ -55,6 +55,9 @@
         '*'.
       - If this value is an IP address, and the C(type) is C(tcp) (the default),
         then a C(port) number must be specified.
+      - In BIG IP Management UI, this field is titled as 'Alias Address'
```
wojtek0806: missing fullstop

library/modules/bigip_monitor_tcp.py
andreykashcheev: Added fullstop as requested

library/modules/bigip_monitor_tcp.py
andreykashcheev: Added fullstop as requested

PR #1708, opened 2020-04-03
Fixes #1275 partially

### LINKED ISSUE #1275: Generate private key and download csr
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
bigip_ssl_key

##### SUMMARY
I would like to generate a private key on the F5 and download the csr to request the certificate to a CA.  
This ca be done in the GUI under System --> File Management --> SSL Certificate List , select Create. Ensure that you choose "Certificate Authority" under Issuer.

nitinthewiz:
There are going to be two parts to this fix. Creating this module is the first part.

### DIFF
--- a/library/modules/bigip_ssl_csr.py
+++ b/library/modules/bigip_ssl_csr.py
@@ -1,0 +1,472 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright: (c) 2019, F5 Networks Inc.
+# GNU General Public License v3.0 (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'certified'}
+
+DOCUMENTATION = r'''
+---
+module: bigip_ssl_csr
+short_description: Create SSL CSR files on the BIG-IP
+description:
+  - This module will create SSL CSR files on a BIG-IP. CSRs
+    require an associated SSL key to pre-exist on the BIG-IP.
+version_added: "f5_modules 1.3"
+options:
+  name:
+    description:
+      - The name of the csr file.
+    type: str
+    required: True
+  common_name:
+    description:
+      - The certificate common name.
+    type: str
+    required: True
+  key_name:
+    description:
+      - The SSL key to be used to generate the CSR.
+    type: str
+    required: True
+  state:
+    description:
+      - When C(present), ensures that the resource exists.
+      - When C(absent), ensures that the resource does not exist.
+    type: str
+    choices:
+      - present
+      - absent
+    default: present
+  dest:
+    description:
+      - Destination on your local filesystem when you want to save the csr file.
+    type: path
+    required: True
+  force:
+    description:
+      - If C(no), the file will only be transferred if the destination does not
+        exist.
+    type: bool
+    default: yes
+extends_documentation_fragment: f5networks.f5_modules.f5
+author:
+  - Nitin Khanna (@nitinthewiz)
+'''
+
+EXAMPLES = r'''
+- name: Create an SSL csr
+  bigip_ssl_csr:
+    name: csr-name
+    key_name: key-name
+    common_name: csr-name
+    dest: /tmp/csr-name
+    provider:
+      password: secret
+      server: lb.mydomain.com
+      user: admin
+  delegate_to: localhost
+'''
+
+RETURN = r'''
+csr_name:
+  description: The name of the CSR file that the user provided
+  returned: created
+  type: str
+  sample: csr-name
+common_name:
+  description: The common name of the CSR file that the user provided
+  returned: created
+  type: str
+  sample: csr-name
+'''
+
+import os
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import env_fallback
+from distutils.version import LooseVersion
+
+try:
+    from library.module_utils.network.f5.bigip import F5RestClient
+    from library.module_utils.network.f5.common import F5ModuleError
+    from library.module_utils.network.f5.common import AnsibleF5Parameters
+    from library.module_utils.network.f5.common import fq_name
+    from library.module_utils.network.f5.icontrol import tmos_version
+    from library.module_utils.network.f5.common import f5_argument_spec
+    from library.module_utils.network.f5.icontrol import download_file
+except ImportError:
+    from ansible_collections.f5networks.f5_modules.plugins.module_utils.bigip import F5RestClient
+    from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import F5ModuleError
+    from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import AnsibleF5Parameters
+    from ansible_collections.f5networks.f5_modules.plugins.module_utils.common import fq_name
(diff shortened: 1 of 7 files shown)


nitinthewiz:
@wojtek0806 I need some guidance on how to run Sanity tests locally so I don't keep running into this issue. Can we talk about this in the team meeting tomorrow? 

Also, this fixes the issue partially with a new module.

wojtek0806:
remove `_bigip_appsvcs_extension.py` from your branch and push it

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+  - This module will create SSL CSR files on a BIG-IP. CSRs
+    require an associated SSL key to prexist on the BIG-IP.
+version_added: "f5_modules 1.2"
+options:
+  name:
+    description:
+      - The name of the csr file.
+    type: str
+    required: True
+  commonName:
```
wojtek0806, line 29: wrong parameter name, no camel case

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+                    'supported_by': 'certified'}
+
+DOCUMENTATION = r'''
+---
+module: bigip_ssl_csr
+short_description: Create SSL CSR files on the BIG-IP
+description:
+  - This module will create SSL CSR files on a BIG-IP. CSRs
+    require an associated SSL key to prexist on the BIG-IP.
+version_added: "f5_modules 1.2"
```
wojtek0806, line 22: wrong version added

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+author:
+  - Nitin Khanna (@nitinthewiz)
+'''
+
+EXAMPLES = r'''
+- name: Create an SSL csr
+  bigip_ssl_csr:
+    name: csr-name
+    key: key-name
+    commonName: csr-name
```
wojtek0806, line 58: wrong parameter

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+      user: admin
+  delegate_to: localhost
+'''
+
+RETURN = r'''
+csr_name:
+  description: The name of the CSR file that the user provided
+  returned: created
+  type: str
+  sample: csr-name
```
wojtek0806, line 71: where is the other returned parameter

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+                continue
+            else:
+                if isinstance(change, dict):
+                    changed.update(change)
+                else:
+                    changed[k] = change
+        if changed:
+            self.changes = UsableChanges(params=changed)
+            return True
+        return False
```
wojtek0806, line 197: if this module cannot update, just create and remove then this is not needed

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+        )
+        resp = self.client.api.patch(uri, json=params)
+        try:
+            response = resp.json()
+        except ValueError as ex:
+            raise F5ModuleError(str(ex))
+
+        if resp.status in [200, 201] or 'code' in response and response['code'] in [200, 201]:
+            return True
+        raise F5ModuleError(resp.content)
```
wojtek0806, line 335: again if you do not update then this is redundant

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+        )
+        resp = self.client.api.get(uri)
+        try:
+            response = resp.json()
+        except ValueError as ex:
+            raise F5ModuleError(str(ex))
+
+        if resp.status in [200, 201] or 'code' in response and response['code'] in [200, 201]:
+            return ApiParameters(params=response)
+        raise F5ModuleError(resp.content)
```
wojtek0806, line 363: we are not updating anything? if so then remove

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+        return False
+
+    def update(self):
+        self.have = self.read_current_from_device()
+        if not self.should_update():
+            return False
+        if self.module.check_mode:
+            return True
+        self.update_on_device()
+        return True
```
wojtek0806, line 259: if not updating anything remove

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+        if LooseVersion(version) < LooseVersion('14.0.0'):
+            return True
+        else:
+            return False
+
+    def present(self):
+        if self.exists():
+            return self.update()
+        else:
+            return self.create()
```
wojtek0806, line 239: if we don't update then just return False on exists() not self.update()

test/integration/bigip_ssl_csr.yaml
```diff
@@ -0,0 +1,52 @@
...
+#
+
+- name: Metadata of bigip_ssl_csr
+  hosts: localhost
+  connection: local
+  gather_facts: false
+
+  vars:
+    __metadata__:
+      upstream: no
```
wojtek0806, line 29: change it to: yes

library/modules/bigip_ssl_csr.py
```diff
@@ -0,0 +1,406 @@
...
+                    'status': ['preview'],
+                    'supported_by': 'certified'}
+
+DOCUMENTATION = r'''
+---
+module: bigip_ssl_csr
+short_description: Create SSL CSR files on the BIG-IP
+description:
+  - This module will create SSL CSR files on a BIG-IP. CSRs
+    require an associated SSL key to prexist on the BIG-IP.
```
wojtek0806, line 21: should be: preexist

library/modules/bigip_ssl_csr.py
nitinthewiz: Gah! I wrote all the testing correctly, just made this documentation mistake! Fixed.

library/modules/bigip_ssl_csr.py
nitinthewiz: fixed

library/modules/bigip_ssl_csr.py
nitinthewiz: fixed

library/modules/bigip_ssl_csr.py
nitinthewiz: Good point. Removed this code.

library/modules/bigip_ssl_csr.py
nitinthewiz: Good point. Removed this code.

library/modules/bigip_ssl_csr.py
nitinthewiz: yeah I set it to return False

nitinthewiz:
> Looking at the module, and thinking more about it, we could add that key upload/download here this means we don't mess with ssl_key module, what is required to have that functionality?

The ask is not for key upload/download. It's for key creation, which should rest in ssl_key if we decide to add it in. Key upload already exists there. What would be required to add that would be simple - a key creation API endpoint exists at sys/crypto/key

wojtek0806:
i have already clarified this with Forrest what we need. I need you to add CSR file download, look at qkview module for a solution on how to achieve that.

nitinthewiz:
> @nitinthewiz
> i have already clarified this with Forrest what we need. I need you to add CSR file download, look at qkview module for a solution on how to achieve that.

I've added the download code based on the qkview module.

Issue #1723, opened 2020-04-15
bigip_profile_client_ssl with no chain defined fails - reintroduced issue #398 


VeselaHouba:
Looks like changes made in https://github.com/F5Networks/f5-ansible/commit/6bb710247e0cb5b141691cc5f58d39a1e5129629 

didn't make it to latest version.

https://github.com/F5Networks/f5-ansible/blob/54d7c0156ea81d9658e5d2251a6917d04feb71d3/library/modules/bigip_profile_client_ssl.py#L595

Chain has to be explicitly set to `none`
Following works:
```
  - bigip_profile_client_ssl:
    name: some_name
    parent: clientssl
    cert_key_chain:
      - cert: cert.crt
        key: key.key
        chain: none
```

Following does not work

```
  - bigip_profile_client_ssl:
    name: some_name
    parent: clientssl
    cert_key_chain:
      - cert: cert.crt
        key: key.key
```

```
fatal: [f501 -> localhost]: FAILED! => changed=false
  msg: one or more configuration identifiers must be provided
```

```
The full traceback is:
  File "/tmp/ansible_bigip_profile_client_ssl_payload_yha8oe8y/ansible_bigip_profile_client_ssl_payload.zip/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_client_ssl.py", line 1269, in main
  File "/tmp/ansible_bigip_profile_client_ssl_payload_yha8oe8y/ansible_bigip_profile_client_ssl_payload.zip/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_client_ssl.py", line 1019, in exec_module
  File "/tmp/ansible_bigip_profile_client_ssl_payload_yha8oe8y/ansible_bigip_profile_client_ssl_payload.zip/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_client_ssl.py", line 1040, in present
  File "/tmp/ansible_bigip_profile_client_ssl_payload_yha8oe8y/ansible_bigip_profile_client_ssl_payload.zip/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_client_ssl.py", line 1075, in update
  File "/tmp/ansible_bigip_profile_client_ssl_payload_yha8oe8y/ansible_bigip_profile_client_ssl_payload.zip/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_client_ssl.py", line 1128, in update_on_device
```

[...]

PR #1729, opened 2020-04-22
fix for #1723

VeselaHouba:
`chain` gets filled to None value by ansible if not defined.
```
    name: some_cert
    parent: clientssl
    cert_key_chain:
      - cert: some_cert.crt
        key: some_cert.key
```
Produces:
```
    {'cert': some_cert.crt', 'key': 'some_cert.key', 'true_names': False, 'chain': None, 'passphrase': None}
```


wojtek0806:
we do not accept external PRs we will fix this in our next release

Issue #1105, opened 2018-12-05
bigip_provision sometimes returns too quickly


c41m4n:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

bigip_provision
##### ANSIBLE VERSION

```
$ [0] ansible --version
ansible 2.7.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/home/exampleuser/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.5 (default, May 31 2018, 09:41:32) [GCC 4.8.5 20150623 (Red Hat 4.8.5-28)]

```

##### PYTHON VERSION

```
$ [0] python -V
Python 2.7.5

```

##### BIGIP VERSION

```
ansible@(lb-fw-m22-test)(cfg-sync Standalone)(/S1-green-P::Active)(/Common)(tmos)# show sys version

Sys::Version
Main Package
  Product     BIG-IP
  Version     13.1.1.2
  Build       0.0.4
  Edition     Point Release 2
  Date        Thu Oct 11 15:32:21 PDT 2018

```

##### LIBRARY VERSIONS

```
 $ [0] pip freeze | egrep '(f5-sdk|bigsuds)'
bigsuds==1.0.6
f5-sdk==3.0.16
```

##### CONFIGURATION

$ [0] cat ansible-datacenter-dev/ansible.cfg
[defaults]
stdout_callback = yaml
bin_ansible_callbacks = True
callback_whitelist = yaml
host_key_checking = false

##### OS / ENVIRONMENT

$ [0] cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.5 (Maipo)

##### SUMMARY

When provisioning modules, I'm finding that about 50% of the time the very next task fails.  I believe the Big-IP hasn't come fully online before the provisioning module completes.
##### STEPS TO REPRODUCE

```
---
- name: Provision F5 Modules
  hosts: f5-test
  connection: local

  tasks:
  - name: f5-provision-modules
    bigip_provision:
      provider: "{{ f5provider }}"
      level: "{{ item.level }}"
      module: "{{ item.name }}"
    delegate_to: localhost
    with_items:
      - "{{ f5_modules }}"

...

f5_modules:
- name: afm
  level: nominal
- name: avr
  level: nominal
- name: asm
[...]

PR #1771, opened 2020-06-16
possible fix for #1105

nitinthewiz:
Still considering what type of testing to write. If a test for VCMP guests already exists, then that's all that's needed


nitinthewiz:
> Still considering what type of testing to write. If a test for VCMP guests already exists, then that's all that's needed

I ran all the currently available testing and observed that the tests were failing for VCMP guests initially and passing after the fix.

PR #1773, opened 2020-06-18
fixed failing bigip_gtm_wide_ip when adding irules to WIP with no irules

D3DeFi:
Module `bigip_gtm_wide_ip` failing when you try to update existing WIP (that has no irules assigned so far) with irules:

```yaml
- name: Create empty WIP
  bigip_gtm_wide_ip:
    name: my-wide-ip.example.com
    provider:
      user: admin
      password: secret
      server: lb.mydomain.com
  delegate_to: localhost

- name: Update empty WIP with irules
  bigip_gtm_wide_ip:
    name: my-wide-ip.example.com
    irules:
      - irule1
    provider:
      user: admin
      password: secret
      server: lb.mydomain.com
  delegate_to: localhost
```

Fix is the same as in `aliases()` method a few lines below this one :)

### DIFF
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_gtm_wide_ip.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_gtm_wide_ip.py
@@ -647,6 +647,8 @@ class Difference(object):
             return None
         if self.want.irules == '' and len(self.have.irules) > 0:
             return []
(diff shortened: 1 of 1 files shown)


wojtek0806:
please read the readme, we do not accept external contributors anymore, closing this PR, we can fix this in next release, for the future just open an issue

PR #1816, opened 2020-08-11
bigip_device_license add renew option

andreykashcheev:
### Summary 

This change is intended to include new option which will be used for license renewal

### Changes
 
 - Update bigip_device_license module to include new option
 - Update unit tests for bigip_device_license module
 - Update integration tests for bigip_device_license module

### Additional details

### DIFF
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_license.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_license.py
@@ -23,7 +23,7 @@ options:
   license_key:
     description:
       - The registration key to use to license the BIG-IP.
-      - This parameter is required if the C(state) is equal to C(present).
+      - This parameter is required if the C(state) is equal to C(present) or C(latest).
       - This parameter is not required when C(state) is C(absent) and will be
         ignored if it is provided.
     type: str
@@ -39,7 +39,7 @@ options:
   license_server:
     description:
       - The F5 license server to use when getting a license and validating a dossier.
-      - This parameter is required if the C(state) is equal to C(present).
+      - This parameter is required if the C(state) is equal to C(present) or C(latest).
       - This parameter is not required when C(state) is C(absent) and will be
         ignored if it is provided.
     type: str
@@ -49,11 +49,13 @@ options:
       - The state of the license on the system.
       - When C(present), only guarantees that a license is there.
       - When C(absent), removes the license on the system.
+      - When C(latest), ensures that the license is always valid. This is not idempotent state since re-run can modify result.
       - When C(revoked), removes the license on the system and revokes its future usage
         on the F5 license servers.
     type: str
     choices:
       - absent
+      - latest
       - present
       - revoked
     default: present
@@ -68,10 +70,18 @@ options:
         ignored if it is provided.
     type: bool
     default: no
+  force:
+    description:
+      - Declares whether to force license renewal. By default, this
+        value is C(no).
+      - This parameter is not required and will be ignored if it is provided.
+    type: bool
+    default: no
 extends_documentation_fragment: f5networks.f5_modules.f5
 author:
   - Tim Rupp (@caphrim007)
   - Wojciech Wypior (@wojtek0806)
+  - Andrey Kashcheev (@andreykashcheev)
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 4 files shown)


test/integration/targets/bigip_device_license/tasks/issue-1810.yaml
```diff
@@ -0,0 +1,29 @@
...
+      license_server: "{{ license_server }}"
+      accept_eula: yes
+      state: "latest"
+  register: result
+
+- name: Assert Try to license device  - Idempotent check
+  assert:
+      that:
+          - result is success
+          - result is not changed
```
wojtek0806, line 29: If this supposed to be not idempotent why this check?

ansible_collections/f5networks/f5_modules/tests/unit/modules/network/f5/test_bigip_device_license.py
```diff
@@ -102,3 +105,75 @@ def test_create(self, *args):
...
+        mm.license_valid = Mock(return_value=True)
+        mm.read_dossier_from_device = Mock(return_value=True)
+        mm.generate_license_from_remote = Mock(return_value=True)
+        mm.upload_license_to_device = Mock(return_value=True)
+        mm.upload_eula_to_device = Mock(return_value=True)
+        mm.reload_license = Mock(return_value=True)
+        mm._is_mcpd_ready_on_device = Mock(return_value=True)
+
+        results = mm.exec_module()
+        assert results['changed'] is False
```
wojtek0806, line 179: I was under the impression you would use the jsons to simulate responses from device and test that way but ok

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_license.py
```diff
@@ -49,11 +49,13 @@
       - The state of the license on the system.
       - When C(present), only guarantees that a license is there.
       - When C(absent), removes the license on the system.
+      - When C(latest), ensures that the license is always valid. This is not idempotent state since re-run can modify result
```
wojtek0806, line 52: fullstop missing

test/integration/targets/bigip_device_license/tasks/issue-1810.yaml
andreykashcheev: I have removed the Idempotent check. Essentially, I kept it there because the state is Idempotent when license is valid  and it would update/change only when license is no longer valid. 

BTW, I also included new "force_renewal" parameter intended to trigger renewal even when license is still valid. 

Let me know if you have any questions/concerns

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_license.py
andreykashcheev: I have added fullstop. This comment is addressed

ansible_collections/f5networks/f5_modules/tests/unit/modules/network/f5/test_bigip_device_license.py
andreykashcheev: Yes - I also thought to use JSON payloads for mocking HTTP response from BIGIP REST API  (i.e. in JS, we use nock module for this: https://github.com/nock/nock)
however it appears to be impossible to do such testing with the current setup.

wojtek0806: it is not impossible look at other unit tests you will see how it is done

test/integration/targets/bigip_device_license/tasks/issue-1810.yaml
wojtek0806: force is enough

andreykashcheev: I have renamed the new parameter to "force"

ansible_collections/f5networks/f5_modules/tests/unit/modules/network/f5/test_bigip_device_license.py
wojtek0806: you can patch F5RestClient, if you are not very intimate with Mock it could be tricky as you need to mock the client instance and then its 'api' property mock to return another mock (the session_mock) and then within the session_mock you return value of .get, but ok lets leave it for now.

PR #1819, opened 2020-08-13
Update bigip_apm_policy_import.py

ajgerace:
Add ability to import API Protection policies


focrensh:
Thank you for the enhancement and PR. We are tracking this with an issue currently referenced above and cannot currently accept the PR directly.

Issue #1880, opened 2020-10-12
bigip_policy_rule: request options in reset action


marcinwito:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

bigip_policy_rule
##### ANSIBLE VERSION

```
ansible-playbook 2.10.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /bin/ansible-playbook
  python version = 2.7.5 (default, Apr  2 2020, 13:16:51) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
```

##### PYTHON VERSION

```
Python 2.7.5
```

##### BIGIP VERSION

```
Version     15.1.0.4
```

##### OS / ENVIRONMENT

N/A
##### SUMMARY

The request option is missing in the event attribute when creating a reset rule for all_traffic.
Only the 'ssl client hello' option is available.
##### STEPS TO REPRODUCE

```
- name: Add cleanup Rule
  bigip_policy_rule:
    policy: "{{ f5_ltm_policy_name }}"
    name: CleanupRule
    conditions:
      - type: all_traffic
    actions:
      - type: reset
        event: request
    provider: "{{ provider }}"
  delegate_to: localhost
```

##### EXPECTED RESULTS

Created rule

Conditions: All traffic
Action: Reset traffic connection at request time
##### ACTUAL RESULTS

```
The error was: KeyError: 'shutdown'
```

nmenant:
We will track this request internally with FMFA-644. Redundant ask from Github#1877

vidarno:
Any ETA on this? 3 months since the issue was raised. We have had to manually implement the fix proposed by the user in #1881 (https://github.com/F5Networks/f5-ansible/pull/1637/commits/83a4934b08284439c78c4e5bf3077a716b5db7c2) which works fine.

PR #1881, opened 2020-10-12
Request options when type is reset

marcinwito:
Add request options to reset type in bigip_policy_rule.  
Resolve https://github.com/F5Networks/f5-ansible/issues/1880


wojtek0806:
thank you for your submission, as per our readme file we no longer accept external pull requests.

Issue #1960, opened 2021-02-01
bigip_device_info.py ansible module does not support 'any` as a validation IP address


kjsanjay:
##### ISSUE TYPE

 - Bug Report
 - F5 ansible module does not support `any` as a valid destination IP address

##### COMPONENT NAME
```
bigip_device_info
```

##### ANSIBLE VERSION
```
ansible 2.10.4
  config file = None
  configured module search path = ['/Users/skayarat/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.10.5/libexec/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.1 (default, Jan  8 2021, 17:15:36) [Clang 11.0.0 (clang-1100.0.33.17)]
```

##### PYTHON VERSION
```
Python 3.6.10

```

##### BIGIP VERSION
```
Sys::Version
Main Package
  Product     BIG-IP
  Version     12.1.3.7
  Build       0.0.2
  Edition     Point Release 7
  Date        Tue Sep 11 10:03:49 PDT 2018
```

##### STEPS TO REPRODUCE

```
tasks:
    - name: Gather device data
      bigip_device_info:
        provider: "{{ provider }}"
        gather_subset:
          - client-ssl-profiles
          - devices
          - device-groups
          - ltm-pools
          - monitors
          - partitions
          - self-ips
          - ssl-certs
          - ssl-keys
          - virtual-servers
      register: gathered_data
      delegate_to: "{{ bastion_host }}"
```

##### EXPECTED RESULTS
```
TASK [Gather device data] ***************************************************************************************************************************************************************
[WARNING]: Platform linux on host [device address redacted] is using the discovered Python interpreter at /usr/bin/python, but future installation of another Python
interpreter could change the meaning of that path. See https://docs.ansible.com/ansible/2.10/reference_appendices/interpreter_discovery.html for more information.
ok: [[device address redacted] -> [device IP redacted]]
```

##### ACTUAL RESULTS

```
TASK [Gather device data] ***************************************************************************************************************************************************************
[WARNING]: Platform linux on host [device address redacted] is using the discovered Python interpreter at /usr/bin/python, but future installation of another Python
interpreter could change the meaning of that path. See https://docs.ansible.com/ansible/2.10/reference_appendices/interpreter_discovery.html for more information.
[...]

PR #1961, opened 2021-02-02
Treating IP address  `any` as a valid destination

kjsanjay:
Some of the documentation suggests that `any` can be used as a valid destination IP address for LTM virtual servers. I can see existing configs on F5 devices that are working fine with `any` as a valid way to specify any destination. But this doesn't seem to be implemented in the ansible module. 

https://clouddocs.f5.com/cli/tmsh-reference/v14/modules/ltm/ltm-virtual.html
```
destination
	    Specifies the name of the virtual address and service on which the
	    virtual server listens for connections.

	    The format for "ipv4" is a.b.c.d[:port]. The format for an "ipv6"
	    address is a:b:c:d:e:f:g:h[.port].

	    The default value is any:any.
```

### DIFF
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_info.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_info.py
@@ -16488,8 +16488,11 @@ class VirtualServersParameters(BaseParameters):
         # 1.1.1.1%2:80
         # /Common/1.1.1.1%2:80
         # /Common/2700:bc00:1f10:101::6%2.any
+        # /Common/any%2:80
+        # /Common/any%2.any
+        # /Common/any%2.80
         #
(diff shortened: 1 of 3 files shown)


ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_info.py
```diff
@@ -16500,7 +16503,7 @@ def destination_tuple(self):
                 if port == 'any':
                     port = 0
             ip = matches.group('ip')
-            if not is_valid_ip(ip):
+            if not is_valid_ip(ip) and ip.lower() != 'any':
```
cvicente, line 16506: How about changing `is_valid_ip()` to support `any` instead?

wojtek0806:
open an issue if you see defects, we do not accept PRs at this time

kjsanjay:
> open an issue if you see defects, we do not accept PRs at this time

https://github.com/F5Networks/f5-ansible/issues/1960

PR #1999, opened 2021-03-26
solving sip profile url issue

GaddamSaketha:


wojtek0806 (changes requested):
have you tested this change against v12 and v13 units? see under tests/integration for plays that create SIP and run this against different versions since this is a bigger change and I got the feeling older BIG-IP versions won't have message routing, since message routing was introduced i believe in v14.

wojtek0806:
@GaddamSaketha  See my comments

wojtek0806:
@GaddamSaketha how are you getting on with this?

wojtek0806:
@GaddamSaketha 

I was looking at it further, how about you create a legacy function for the legacy SIP profiles in addition to your change, add the new function here and it should be ok we wont have any regression concerns:

https://github.com/F5Networks/f5-ansible/blob/64b797be79c41cb5ae7a7e72386490ae21558fdc/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_virtual_server.py#L1182

name the legacy function something like:

self._read_legacy_sip_profiles_from_device()

Issue #1978, opened 2021-03-04
bigip_asm_import_policy: policy with name being substring of existing policy won't be imported


amolari:
##### ISSUE TYPE

 - Bug Report
 
##### COMPONENT NAME

bigip_asm_import_policy
##### ANSIBLE VERSION

```
2.9.15
```

##### PYTHON VERSION

```
python version = 3.6.8 (default, Aug 24 2020, 17:57:11) [GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]
```

##### BIGIP VERSION

```
14.1.3.1
```

##### SUMMARY

I have an already installed policy with the name "dev-event-declaration.example.com"
I want to install a new policy with name "event-declaration.example.com"
The policy won't be installed because no change has been detected.
Issue is most probably here:
https://github.com/F5Networks/f5-ansible/blob/79bb5ac8e4757b77eba60556d80bad82e3001bf1/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_asm_policy_import.py#L433-L435

because name "event-declaration.example.com" is a substring of "dev-event-declaration.example.com".
It should be an equal, not contains.

##### STEPS TO REPRODUCE

```

```

##### EXPECTED RESULTS

Policy event-declaration.example.com can be installed
##### ACTUAL RESULTS

No change detected, policy won't be installed

```

```

amolari:
changed
`query = "?$filter=contains(name,'{0}')+and+contains(partition,'{1}')&$select=name,partition".format( `
to
`query = "?$filter=name+eq+'{0}'+and+contains(partition,'{1}')&$select=name,partition".format(`
worked for me. But probably the same should be done with the partition

trinaths:
created JIRA INFRAANO-323 for internal PM tracking

wojtek0806:
fixed with #2042

PR #2042, opened 2021-05-31
multiple_fixes

wojtek0806:
fix for #1978


DerrickTJ:
Sweet! Any idea what version of Ansible this will end up shipping with?

wojtek0806:
collection 1.10

PR #2059, opened 2021-06-24
fix for API calls to get current cert and key path on bigip

fairytaletrash:
API response in 16.0.1 returns fullPath instead of sourcePath for /mgmt/tm/sys/file/ssl-key/ and /mgmt/tm/sys/file/ssl-cert/ API requests.  This breaks the modules ability to replace an existing cert/key pair.  Change is to reference correct key in JSON from API response


trinaths:
Unable to consider this PR. Unit tests and change log are missing. This PR requires more testing before merging. Closing this PR.

PR #2068, opened 2021-07-06
fix software install issue if volume exist

RavinderReddyF5:


wojtek0806 (changes requested):
Add functional test for this
add changelog
volume removal should be a separate function

Issue #1912, opened 2020-11-13
bigip_profile_fastl4 parameter module syn-cookie-enable


gitnetofr:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
```
bigip_profile_fastl4
```
##### ANSIBLE VERSION
```
ansible 2.10.1 
collection 1.6.0
```
##### PYTHON VERSION
```
3.8
```

##### BIGIP VERSION
```
Version     15.1.0.5
Build       0.0.8
BIG-IP VE
```
##### CONFIGURATION
```
# default values
[defaults]
inventory                     = ./inventory/
roles_path                    = ./roles/
callback_whitelist            = timer
host_key_checking             = False
force_color                   = 1
ansible_python_interpreter    =/usr/bin/python3
[ssh_connection]
ssh_args = -o UserKnownHostsFile=/dev/null -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s -o StrictHostKeyChecking=no 
```

##### OS / ENVIRONMENT

```
Ubuntu 20.04
```

##### SUMMARY
Request a ansible paramter for module bigip_profile_fastl4.
I can't find the option 'SYN Challenge Handling'. It seems that it is not available.
The tmsh call for the parameter is "tmsh modify ltm profile fastl4 profile-fastL4-name syn-cookie-enable disabled"
Could you add this parameter in the roadmap developpment?

##### STEPS TO REPRODUCE
I want to write a playbook which creates a fastL4 profile configured like that:
tmsh create ltm profile fastl4 profile-fastL4-name defaults-from fastL4 reset-on-timeout enabled loose-close enabled loose-initialization enabled syn-cookie-enable disabled
My playbook is blocked because the parameter syn-cookie-enable is not available.

##### EXPECTED RESULTS
```
root@(opntrrpv-s5014)(cfg-sync Standalone)(Active)(/Common)(tmos)# list ltm profile fastl4 profile-fastL4-name                                                                         ltm profile fastl4 profile-fastL4-name {
    app-service none
    defaults-from fastL4
    loose-close enabled
    loose-initialization enabled
    reset-on-timeout enabled
    syn-cookie-enable disabled
}
```

onurbrc:
Hi, 

I'm also need this option. Any news? 

Thanks in advance.

focrensh:
Tracking via INFRAANO-40

onurbrc:
Hi, 

Any news about this parameter? 

[...]

PR #2078, opened 2021-07-20
add syn_cookie_enable parameter to bigip_profile_fastl4 module

papineni87:
Fix for #1912

### DIFF
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_fastl4.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_fastl4.py
@@ -286,7 +286,7 @@ options:
     type: bool
   syn_cookie_enable:
     description:
-      - Specifies whether or not to use SYN Cookie. 
+      - Specifies whether or not to use SYN Cookie.
     type: bool
   pva_acceleration:
     description:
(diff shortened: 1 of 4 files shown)


ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_fastl4.py
```diff
@@ -284,6 +284,10 @@
     description:
       - Enables or disables hardware SYN cookie support when PVA10 is present on the system.
     type: bool
+  syn_cookie_enable:
+    description:
+      - Specifies whether or not to use SYN Cookie.
+    type: bool
```
wojtek0806, line 290: missing version_added

wojtek0806 (changes requested):

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_fastl4.py
papineni87: Added version

wojtek0806: this needs to be "1.11.0" as its new feature only bugs increment the last digit, which is why most likely you have sanity error, also i think your indentation is a bit off recheck it

wojtek0806 (changes requested):

wojtek0806 (approved):

Issue #2071, opened 2021-07-09
validate_ip_address does not allow for IP addresses with %RD notation


MitchellMahan:
##### COMPONENT NAME
validate_ip_address()

##### SUMMARY
When creating a virtual_address I would like to use 1.1.1.1%13 to specify route domain 13.

Because the validate_ip_address function does not allow for %n notation all functions that use validate_ip_address fail when specifying a route domain.

##### STEPS TO REPRODUCE
Specify a %n route domain when using bigip_virtual_server, bigip_virtual_address, etc. etc.

##### EXPECTED RESULTS
1.1.1.1%13 should be an allowable IP address.

validate_ip_address(ip):
     # Use proper regex validation (maybe the same regex as the F5 NSO YANG?)
    pass

trinaths:
@MitchellMahan - This can be easily achieved with Collections V2. Refer to [documentation](https://clouddocs.f5.com/products/orchestration/ansible/devel/f5_bigip/install_f5_bigip.html) to install Collections V2.

Sample AS3 JSON and playbook below:

```
{
  "class": "AS3",
  "action": "deploy",
  "persist": true,
  "declaration": {
    "class": "ADC",
    "schemaVersion": "3.22.0",
    "id": "id",
    "label": "WebApp",
    "ansible": {
      "class": "Tenant",
      "A1": {
        "class": "Application",
        "web": {
          "class": "Service_HTTP",
          "virtualAddresses": ["10.1.2.3%33"],             #<--- VIP with RouteDomain
          "virtualPort": 80,
          "pool": "web_pool"
        },
        "web_pool": {
          "class": "Pool",
          "monitors": [{"use": "http_mon"}],
          "members": [
            {
              "servicePort": 8080,
              "serverAddresses": [
                "10.2.2.3"
              ]
            }
          ]
        },
        "http_mon": {
            "class": "Monitor",
            "monitorType": "http",
            "receive": "I AM UP"
        }
      }
    }
  }
}
```

```
---
- name: AS3
  hosts: bigip
  connection: httpapi
  gather_facts: false

  # Connection Info
  vars:
    ansible_host: <bigip-host>
    ansible_user: admin
    ansible_httpapi_password: <some-password>
    ansible_httpapi_port: 443
    ansible_network_os: f5networks.f5_bigip.bigip
    ansible_httpapi_use_ssl: yes
    ansible_httpapi_validate_certs: no

  tasks:

    - name: Deploy or Update AS3
[...]

PR #2079, opened 2021-07-20
Allow users to specify ip_address%rd in f5_modules

MitchellMahan:
This pull request is an example of the fix necessary for issue #2071

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,28 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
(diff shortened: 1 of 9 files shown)


wojtek0806:
Before I attempt to review this here are things right of the bad:

proper changelog fragment is missing
no functional tests made (see integration tests for examples), also if you make change like that not only your tests but all tests for the affected modules need to be run so we ensure there are no issues (we post the test results here)
you must not edit galaxy.yml file

wojtek0806:
closing this PR as the functionality to specify RD in Virtual address and virutal server exists already. If you see issues with RD in other modules then raise them as separate issues.

Issue #1473, opened 2019-08-28
 --diff does not produce diff


erikanderson:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME

All
##### ANSIBLE VERSION

```
ansible 2.8.4
  config file = /Users/eanderson/code/ansible/ansible.cfg
  configured module search path = ['/Users/eanderson/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.7.4 (default, Jul  9 2019, 18:13:23) [Clang 10.0.1 (clang-1001.0.46.4)]
```

##### PYTHON VERSION

```
Python 3.7.4
```

##### BIGIP VERSION

```
Sys::Version
Main Package
  Product     BIG-IP
  Version     12.1.3.5
  Build       0.0.10
  Edition     Point Release 5
  Date        Mon May 21 17:36:20 PDT 2018
```

##### ROLE VERSION

```
Role: f5devcentral.f5ansible
        description: Ansible role that allows use of the F5 Ansible Modules
        active: True
        commit: 3c1b42c67583b42718b6916abdc4cd95d8c1c835
        commit_message: Merging upstream changes - 2019-08-24T07:00:31Z
        commit_url: https://api.github.com/repos/f5devcentral/ansible-role-f5ansible/git/commits/3c1b42c67583b42718b6916abdc4cd95d8c1c835
        company: F5 Networks

        created: 2018-10-04T16:14:10.510085Z
        download_count: 12877
        forks_count: 2
        github_branch: master
        github_repo: ansible-role-f5ansible
        github_user: f5devcentral
        id: 30341
        imported: 2019-08-24T03:03:09.617109-04:00
        is_valid: True
        issue_tracker_url:
        license: GPLv3
        min_ansible_version: 2.8
        modified: 2019-08-24T07:03:09.623354Z
        open_issues_count: 0
        path: ('/Users/eanderson/code/ansible/roles',)
        role_type: ANS
        stargazers_count: 4
        travis_status_url: https://travis-ci.org/f5devcentral/ansible-role-f5ansible.svg?branch=master
```

##### OS / ENVIRONMENT
OSX10.14.6

##### SUMMARY
Running module with `--diff` does not produce diff

##### STEPS TO REPRODUCE

```
[...]

PR #2087, opened 2021-08-05
support --diff mode

cawamata:
## Issue
Some modules does not support diff option

## Discussion Point
From #1473, I'm not sure the below is finished.
> this is a deliberate choice not to implement this in most modules as this will be done in stages. Some modules that have aggregates will not have diff implemented until aggregates are reworked completely, these modules are: bigip_pool_member, bigip_pool, bigip_gtm_pool_member.

However, I need --diff mode, then I created this PR.
At first, I'd like to confirm this way is okay or not. After design is fixed, I will add another modules which not support this mode in this PR


cawamata:
@trinaths 
Can you tell me more detail to progress to add tis option?
Should I add sanity test to confirm diff output? or should I add the output difference of with/without this change?
As you know, all users cannot know what point will be updated beforehand, then this point is risk.

PR #2129, opened 2021-11-11
fix for source capability for bigip_device_auth_ldap module

RavinderReddyF5:


ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_auth_ldap.py
```diff
@@ -150,6 +150,14 @@
...
         without removing the configured LDAP resource.
     type: bool
+  source_type:
+    description:
+      - Specifies auth source to be used for user authentication,should be used with C(use_for_auth)
+    type: str
+    choices:
+      - ldap
+      - active-directory
+    default: ldap
```
wojtek0806, line 160: `version_added: "1.13.0"`

wojtek0806 (changes requested):
change the file as requested, do we need to have tests here?

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_auth_ldap.py
RavinderReddyF5: updated

RavinderReddyF5: tests also added

wojtek0806 (approved):

PR #2207, opened 2022-04-30
added support for "up_interval" to HTTP monitor module.

rhavenn:
Added support to the HTTP monitor module for the 'up_interval' setting.

I verified this worked against a test instance of 16.1.2.1 we manage.


wojtek0806:
Thank you for your submission, we will include part of this PR in https://github.com/F5Networks/f5-ansible/pull/2273

PR #2247, opened 2022-08-26
Add asm to disable_target list 

anderpups:
Add asm to disable_target list for bigip_policy_rule action disable.

This will allow you to disable asm as part of a policy rule.


wojtek0806:
Thank you for your submission, we will include part of this PR in https://github.com/F5Networks/f5-ansible/pull/2273

Issue #2261, opened 2022-10-13
bigip_device_group: add feature to allow ASM config sync


simonkowallik:
### Is your feature request related to a problem? Please describe.
f5-ansible allows to create device-groups but does not provide the option to set `asm-sync`.

### Describe the solution you'd like
Provide an option to set `asm-sync` in `bigip_device_group`.

### Describe alternatives you've considered
The alternative is to create/modify the device-group via tmsh, possibly via `bigip_command`.

PR #2262, opened 2022-10-13
feature: add asm config sync option

simonkowallik:
This PR adds the option to set `asm-sync` on a device-group, see #2261

### DIFF
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
@@ -54,6 +54,7 @@ options:
       - When creating a new device group, this option will default to C(no).
     type: bool
     default: no
+    version_added: "1.21.0"
   save_on_auto_sync:
     description:
       - When performing an auto-sync, specifies whether the configuration
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
@@ -47,6 +47,13 @@ options:
       - When creating a new device group, this option will default to C(no).
     type: bool
     default: no
+  asm_sync:
+    description:
+      - Specifies whether to synchronize ASM configurations of device group members.
+      - A device can be a member of only one ASM-enabled device group.
+      - When creating a new device group, this option will default to C(no).
+    type: bool
+    default: no
   save_on_auto_sync:
     description:
       - When performing an auto-sync, specifies whether the configuration
@@ -121,6 +128,17 @@ EXAMPLES = r'''
       server: lb.mydomain.com
       user: admin
   delegate_to: localhost
+
+- name: Create a sync-only device group with auto-sync and asm-sync enabled
+  bigip_device_group:
+    name: foo-group
+    auto_sync: yes
+    asm_sync: yes
+    provider:
+      password: secret
+      server: lb.mydomain.com
+      user: admin
+  delegate_to: localhost
 '''
 
 RETURN = r'''
@@ -146,6 +164,11 @@ type:
   sample: sync-failover
 auto_sync:
   description: The new auto_sync value of the device group.
+  returned: changed
+  type: bool
+  sample: true
+asm_sync:
+  description: The new asm_sync value of the device group.
   returned: changed
   type: bool
   sample: true
@@ -177,6 +200,7 @@ class Parameters(AnsibleF5Parameters):
         'saveOnAutoSync': 'save_on_auto_sync',
         'fullLoadOnSync': 'full_sync',
         'autoSync': 'auto_sync',
+        'asmSync': 'asm_sync',
         'incrementalConfigSyncSizeMax': 'max_incremental_sync_size',
         'networkFailover': 'network_failover',
     }
@@ -186,6 +210,7 @@ class Parameters(AnsibleF5Parameters):
         'description',
         'type',
         'autoSync',
+        'asmSync',
         'incrementalConfigSyncSizeMax',
         'networkFailover',
     ]
@@ -195,6 +220,7 @@ class Parameters(AnsibleF5Parameters):
         'description',
         'type',
         'auto_sync',
+        'asm_sync',
         'max_incremental_sync_size',
         'network_failover',
     ]
@@ -203,6 +229,7 @@ class Parameters(AnsibleF5Parameters):
         'full_sync',
         'description',
         'auto_sync',
+        'asm_sync',
         'max_incremental_sync_size',
         'network_failover',
     ]
@@ -239,6 +266,14 @@ class ApiParameters(Parameters):
         if self._values['auto_sync'] is None:
             return None
         elif self._values['auto_sync'] == 'enabled':
+            return True
+        return False
+
+    @property
+    def asm_sync(self):
+        if self._values['asm_sync'] is None:
+            return None
+        elif self._values['asm_sync'] == 'enabled':
             return True
         return False
 
@@ -299,6 +334,14 @@ class UsableChanges(Changes):
         return 'disabled'
 
     @property
+    def asm_sync(self):
+        if self._values['asm_sync'] is None:
+            return None
+        elif self._values['asm_sync']:
+            return 'enabled'
+        return 'disabled'
+
+    @property
     def save_on_auto_sync(self):
         if self._values['save_on_auto_sync'] is None:
             return None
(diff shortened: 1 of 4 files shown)


ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
```diff
@@ -47,6 +47,13 @@
       - When creating a new device group, this option will default to C(no).
     type: bool
     default: no
+  asm_sync:
+    description:
+      - Specifies whether to synchronize ASM configurations of device group members.
+      - A device can be a member of only one ASM-enabled device group.
+      - When creating a new device group, this option will default to C(no).
+    type: bool
+    default: no
```
wojtek0806, line 56: missing version_added parameter

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
```diff
@@ -242,6 +269,14 @@ def auto_sync(self):
...
         return False
 
+    @property
+    def asm_sync(self):
+        if self._values['asm_sync'] is None:
+            return None
+        elif self._values['asm_sync'] == 'enabled':
+            return True
+        return False
+
```
wojtek0806, line 279: remove this its not needed

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
```diff
@@ -298,6 +333,14 @@ def auto_sync(self):
...
         return 'disabled'
 
+    @property
+    def asm_sync(self):
+        if self._values['asm_sync'] is None:
+            return None
+        elif self._values['asm_sync']:
+            return 'enabled'
+        return 'disabled'
+
```
wojtek0806, line 343: use flatten_boolean function here, and do something like
```
result = flatten_boolean(self._values['asm_sync'])
    if result == 'yes':
       return 'enabled
    if result == 'no':
      return 'disabled'

File: ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py [hunk 4]
```diff
@@ -575,6 +626,10 @@ def __init__(self):
                 type='bool',
                 default='no'
             ),
+            asm_sync=dict(
```
wojtek0806, line 629: no defaults , if you want to have create default add it to create stage, also if we created device group before without this option specified then having default for create is pointless, we only add default create arguments for mandatory options that are required for new resources

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
```diff
@@ -334,6 +377,14 @@ def auto_sync(self):
...
         return 'no'
 
+    @property
+    def asm_sync(self):
+        if self._values['asm_sync'] is None:
+            return None
+        elif self._values['asm_sync'] == 'enabled':
+            return 'yes'
+        return 'no'
+
```
wojtek0806, line 387: remove

wojtek0806 (changes requested):
You need to add functional tests and paste their test results, functional tests should contain idempotency tests to verify adding new parameter does not break module's idempotency, have a  look at tests/intergration/targets for some ideas.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
simonkowallik: Will add it, thanks.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
simonkowallik: The `asm_sync` properties is quite similar to `auto_sync` in its behaviour within the API, hence this is required. Removing this leads to the module not being able to understand the API parameters and always starting to report changes for this setting.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
simonkowallik: I've mirrored the approach taken for all other API attributes, specifically `auto_sync`. `flatten_boolean` isn't a function currently used within the code which is why I think it would be better to stick to the existing approach. If this needs to be refactored later this could be done along with all other properties.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
simonkowallik: Again I used `auto_sync` as a working example. Removing this would not restore the BIG-IP default behaviour of not syncing ASM policies via the device-group. For example when `asm_sync: yes` is just removed from the ansible task this setting would not be updated and `asmSync` would stay `enabled` in the API.
The above code mirrors the `auto_sync` behaviour exactly, which happens to be disabled by default as well.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
simonkowallik: Removing this would lead to the module not detecting whether the setting needs to be changed, therefore it is required (again, same as for `auto_sync`).

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
wojtek0806: if the user has this changed to enabled already, then running playbook without this parameter will cause the change, remove it

wojtek0806: we do not use defaults for api parameters only for parameters that are not linked directly to API

wojtek0806 (changes requested):

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
wojtek0806: you are going by wrong examples, flatten boolean is for ensuring that if i specify for example: true, yes, True etc. in the playbook we make it either 'yes' and 'no' same goes for whatever api returns, if we see enabled/on/True etc we ensure internally we have it as 'yes' or 'no' this is done for comparison. Look at other modules how we handle it.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
wojtek0806: you are going by wrong examples, flatten boolean is for ensuring that if i specify for example: true, yes, True etc. in the playbook we make it either 'yes' and 'no' same goes for whatever api returns, if we see enabled/on/True etc we ensure internally we have it as 'yes' or 'no' this is done for comparison. Look at other modules how we handle it.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
wojtek0806: and to be honest we should change it here cause its not very consistent with rest of code

simonkowallik: I see, you are referring to existing configuration conflicting with the defaults in this module. I would still keep the default because the behaviour for `asm_sync` would exactly match the existing behaviour of `auto_sync`. I would also assume that any configuration object created via or added to ansible is _owned_ by ansible and MUST NOT be modified outside ansible.

For example assume the below device group was created outside of ansible:
```hcl
cm device-group my-device-group {
    auto-sync enabled
}
```

Then running the below ansible task:
```yaml
- name: create device-group
  f5networks.f5_modules.bigip_device_group:
    name: 'my-device-group'
  delegate_to: localhost
```
Will result in:
```sh
changed: [bigip -> localhost({{ provider.server }})] => {"auto_sync": "no", "changed": true}
```
```hcl
cm device-group my-device-group { }
```

The same is true for `asm_sync` but it just mirrors existing behaviour. If the existing behaviour is undesired I would rather see this as part for a bigger refactoring of this module than adding new behaviour through this feature addition.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
simonkowallik: Removing this leads to module failures as the API parameter would not be recognized anymore (same behaviour applies as for `auto_sync`/`autoSync`).
Refactoring to `flatten_boolean` seems like a good idea, other modules already use it, but this would be beyond the scope of this feature PR as it applies to all properties of class `ApiParameters`.

I think refactoring `flatten_boolean` should be considered for a separately.

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_group.py
simonkowallik: Refactoring to `flatten_boolean` seems like a good idea, other modules already use it, but this would be beyond the scope of this feature PR as it applies to all properties of class `UsableChanges`.

I think refactoring `flatten_boolean` should be considered for a separately.

wojtek0806:
Thank you for your submission, we will include part of this PR in #2273

PR #2279, opened 2022-11-28
removes distutils package

wojtek0806:
adds package.version as replacement


wojtek0806:
ignoring sanity failures for now, merging as it is

PR #2307, opened 2023-03-02
Handle all non-success results returned from device

caleb-gosnell:
Submitting in the hope that you'll see the value of this straightforward change.

We were seeing a problem with our playbooks where they would fail inexplicably with stack traces from inside the module code with messages like `KeyError: items` because the call to the F5 had failed for some reason unrelated to the validity of the playbook/task parameters (e.g. an auth token expires during the playbook run), and the results from the F5 would not match what was expected. 

This change prevents confusion about the source of such an issue (playbook definition vs F5 itself), which is an important distinction when debugging playbook runs.


wojtek0806:
we won't merge it as it is but we will handle this internally to make required improvements, tracking: INFRAANO-1111

PR #2431, opened 2024-10-23
BGV route domain bugfix in bigip_imish

urohit011:


ansible_collections/f5networks/f5_modules/plugins/modules/bigip_imish_config.py
```diff
@@ -629,18 +630,26 @@ def load_config_on_device(self, name):
...
+                response = resp.json()
+            except ValueError as ex:
+                raise F5ModuleError(str(ex))
+
+            if resp.status in [200, 201] or 'code' in response and response['code'] in [200, 201]:
+                if 'commandResult' in response:
+                    if 'Dynamic routing is not enabled' in response['commandResult']:
+                        raise F5ModuleError(response['commandResult'])
+                    if 'Protocol daemon is not running' in response['commandResult']:
+                        x += 1
```
RavinderReddyF5, line 647: HI @urohit011 ,
do you think 15 seconds of sleep is enough, is there any scenario it takes more time

RavinderReddyF5 (approved):
LGTM

ansible_collections/f5networks/f5_modules/plugins/modules/bigip_imish_config.py
urohit011: so far it seems that 5 seconds are enough.

PR #2441, opened 2024-11-19
Fixes for issue 2438

gomesjj:
Adding the ability to set virtual server option "serverssl-use-sni" using the "bigip-virtual_server module, as well as gathering the option's status with "bigip-device-info".


gomesjj:
Corrected issues at build: removed trailing whitespaces and added blank line to end of files.

urohit011 (approved):

Issue #2454, opened 2025-03-06
bigip_firewall_address_list won't configure address list with both CIDR mask and route domain


jimmylela:
##### COMPONENT NAME
f5networks.f5_modules.bigip_firewall_address_list

##### ANSIBLE VERSION

```
ansible [core 2.13.13]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/<username_redacted>/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /mnt/c/Users/<username_redacted>/scripts/ansible/wsl_ansible/venv/lib/python3.8/site-packages/ansible
  ansible collection location = /home/<username_redacted>/.ansible/collections:/usr/share/ansible/collections
  executable location = /mnt/c/Users/<username_redacted>/scripts/ansible/wsl_ansible/venv/bin/ansible
  python version = 3.8.10 (default, Feb  4 2025, 15:02:54) [GCC 9.4.0]
  jinja version = 3.1.3
  libyaml = True
```
##### BIGIP VERSION

```
Sys::Version
Main Package
  Product     BIG-IP
  Version     17.1.2
  Build       0.0.8
  Edition     Final
  Date        Thu Nov 21 04:35:41 PST 2024
```
##### SUMMARY
I am able to create a new address list with a CIDR subnet or a route domain, but not both. I cant find any documentation on the syntax for both, but many of the ways I've tried run successfully with no errors or warnings and create the address list on the target F5, but the CIDR mask is missing. The equivalent TMSH command "create security shared-objects address-list high_risk_ips_jimmy_test { addresses add { 135.148.10.160%5/28 } }" works as expected.

##### STEPS TO REPRODUCE

```
- name: configure address list on F5
  hosts: <redacted>
  gather_facts: false
  connection: local

  tasks:

  - name: configure address list
    f5networks.f5_modules.bigip_firewall_address_list:
      provider: "{{ provider }}"
      name: high_risk_ips_jimmy_test
      addresses:
        ## some formats I've tried
        # - 135.148.10.160%5/28
        # - "135.148.10.160%5/28"
        # - "135.148.10.160%5\/28"
        # - 135.148.10.160%5\/28
        # - '135.148.10.160%5/28'
        # - '135.148.10.160%5\/28'
        - 135.148.10.160%5/28
```

##### EXPECTED RESULTS
[...]

PR #2457, opened 2025-04-03
Fix for bigip_firewall_address_list to support both CIDR mask and route domain

pmaddev:
Fix for bigip_firewall_address_list to support both CIDR mask and route domain

#2454


pmaddev:
Ran sanity and test cases

inv test.f5-sanity
 
inv test.style
 
inv test.unit
 
inv test.ansible-test -r -p 3.10

RavinderReddyF5 (approved):
LGTM

PR #2460, opened 2025-04-08
Issue with parent argument accepting None but expecting a string in bigip_profile_server_ssl module

### LINKED ISSUE #2459: Issue with parent argument accepting None but expecting a string in bigip_profile_server_ssl module
When passing parent: !!null (which corresponds to None in Ansible) in the bigip_profile_server_ssl module, the module throws an error stating that the parent argument is expected to be a string but received None. The error message is as follows:
{"changed": false, "msg": "argument 'parent' is of type <class 'NoneType'> and we were unable to convert to str: 'None' is not a string and conversion is not allowed"}

The module handles the parent argument in this way in the parent function:

def parent(self):
    if self._values['parent'] is None:
        return None
    if self._values['parent'] == 'serverssl':
        return '/Common/serverssl'
    result = fq_name(self.partition, self._values['parent'])
    return result

The Issue:
The function parent() seems to handle None correctly by returning None when the argument is None. However, when parent is None, the module fails in other parts of the code, likely where a string value is expected. The error message (argument 'parent' is of type <class 'NoneType'> and we were unable to convert to str) indicates that a string was expected, but None was passed.

Steps to Reproduce:
Pass parent: !!null (i.e., None in Ansible) in the playbook for the bigip_profile_server_ssl module.

Run the playbook, and observe the error.

Expected Behavior:
The module should either:

Properly handle the None case in all relevant code paths (e.g., when interacting with an external API that expects a string).

pmaddev:
Issue: Issue with parent argument accepting None but expecting a string in bigip_profile_server_ssl module

Fix: Added support for parent profile None and null

Resolves: #2459 

Ran Sanity and unit test cases

inv test.f5-sanity
inv test.style
inv test.unit
inv test.ansible-test -r -p 3.13

### DIFF
--- a/ansible_collections/f5networks/f5_modules/changelogs/fragments/2458-profile-server-ssl.yaml
+++ b/ansible_collections/f5networks/f5_modules/changelogs/fragments/2458-profile-server-ssl.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - bigip_profile_server_ssl to support parent's [None, "", "None"] profiles
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_server_ssl.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_profile_server_ssl.py
@@ -24,7 +24,7 @@ options:
     description:
       - The parent template of this monitor template. Once this value has
         been set, it cannot be changed.
-    type: str
+    type: raw
     default: /Common/serverssl
   ciphers:
     description:
@@ -383,10 +383,11 @@ class ModuleParameters(Parameters):
 
     @property
     def parent(self):
-        if self._values['parent'] is None:
+        parent = self._values.get("parent")
+        if parent in [None, "", "None"]:
             return None
-        if self._values['parent'] == 'serverssl':
-            return '/Common/serverssl'
+        if parent == "serverssl":
+            return "/Common/serverssl"
         result = fq_name(self.partition, self._values['parent'])
         return result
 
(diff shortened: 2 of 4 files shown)


pmaddev:
inv test.ansible-test -r -p 3.13
Running sanity test "action-plugin-docs"
ERROR: Found 2 action-plugin-docs issue(s) which need to be resolved:
ERROR: plugins/action/bigip.py:0:0: action plugin has no matching module to provide documentation
ERROR: plugins/action/bigiq.py:0:0: action plugin has no matching module to provide documentation
See documentation for help: https://docs.ansible.com/ansible-core/2.18/dev_guide/testing/sanity/action-plugin-docs.html
Running sanity test "ansible-doc"
Running sanity test "changelog"
Running sanity test "compile" on Python 3.13
Running sanity test "empty-init"
Running sanity test "ignores"
Running sanity test "import" on Python 3.13
Running sanity test "line-endings"
Running sanity test "no-assert"
Running sanity test "no-get-exception"
Running sanity test "no-illegal-filenames"
Running sanity test "no-smart-quotes"
Running sanity test "pep8"
Running sanity test "pslint"
Running sanity test "pylint"
Running sanity test "replace-urlopen"
Running sanity test "runtime-metadata"
Running sanity test "shebang"
Running sanity test "shellcheck"
Running sanity test "symlinks"
Running sanity test "use-argspec-type-path"
Running sanity test "use-compat-six"
Running sanity test "validate-modules"
Running sanity test "yamllint"
FATAL: The 1 sanity test(s) listed below (out of 24) failed. See error output above for details.
action-plugin-docs

RavinderReddyF5 (approved):
LGTM

venkateshkalluri:
@RavinderReddyF5 and @pgouband ,
Just wondering, do you have an estimated timeline for the next release? I’d like to know when the fix will be available so we can plan to use the updated modules in our environment.

pgouband:
Hi @venkateshkalluri,

Latest release has the fix implemented.

Issue #2490, opened 2025-09-09
bigip_ike_peer issue creating IKEv2


snuffy22:
##### COMPONENT NAME

bigip_ike_peer

##### ANSIBLE VERSION

```
ansible [core 2.18.7]
  config file = None
  configured module search path = ['/home/support/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/support/f5-env/lib/python3.12/site-packages/ansible
  ansible collection location = /home/support/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/support/f5-env/bin/ansible
  python version = 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] (/home/support/f5-env/bin/python3)
  jinja version = 3.1.6
  libyaml = True

```
##### BIGIP VERSION
```
Sys::Version
Main Package
  Product     BIG-IP
  Version     17.5.0
  Build       0.0.15
  Edition     Final
  Date        Wed Feb 19 02:13:04 PST 2025
```
##### OS / ENVIRONMENT
Ubuntu 24.04 LTS

##### SUMMARY
When creating a IKE v2 entry we need to specify a Traffic Selector during creation, also specifying 'override' for the verification is not accepted had to resort to just using IP

##### STEPS TO REPRODUCE

```
inventory:

all:
  children:
    f5_cluster:
      hosts:
        f5-ve-01:
          f5_hostname: f5-ve-01.local
          f5_peer: f5-ve-02.local
          ansible_host: 192.168.3.150
          peer_device: 192.168.3.157
          primary_device: true
        f5-ve-02:
          f5_hostname: f5-ve-02.local
          f5_peer: f5-ve-01.local
          ansible_host: 192.168.3.157

        vars:
          provider:
            server: "{{ ansible_host }}"
            user:  admin
            password: test123
            validate_certs: false

playbook:
- name: Setup F5 HA Trust and Sync Group
  hosts: f5_cluster
  gather_facts: false
  connection: local
  collections:
    - f5networks.f5_modules
   
  tasks: 
      - name: Create a IPSec policy
        bigip_ipsec_policy:
          provider: "{{ provider }}"
          name: test_ipsec
          mode: tunnel
          tunnel_local_address: 192.168.60.1
          tunnel_remote_address: 192.168.60.5
          auth_algorithm: aes256
          encrypt_algorithm: sha256
          protocol: esp
[...]

PR #2491, opened 2025-09-09
Update bigip_ike_peer.py to support traffic selector ikev2

snuffy22:
Allows for traffic selector to be specified so IKEv2 peer can be created


snuffy22:
This PR adds traffic selector for IKEv2 peers
Fixes #2490

Issue #2509, opened 2026-01-18
bigip_virtual_server.py - Bug with metadata inconsistently set (removed each 1 of 2 run)


amolari:
##### COMPONENT NAME
bigip_virtual_server.py

##### ANSIBLE VERSION

```
2.16.9
```
##### BIGIP VERSION

```
17.5.1.3
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

N/A
##### STEPS TO REPRODUCE

I create a VS with metadata being a var (dictionary)
```
- name: set fact vs_metadata
  set_fact:
    vs_metadata:
      run_caller: "{{ run_caller }}"
      application: "{{ application | regex_replace('^[A-Z-]+_(.*)$', '\\1') }}"

- name: set fact vs_metadata - add conditionally httpver=2 (when http2_enabled)
  set_fact:
    vs_metadata: "{{ vs_metadata | combine({ 'httpver': '2' }) }}"
  when: http2_enabled

- name: Add VS "{{ application }}_vs" - with config map
  bigip_virtual_server:
    provider: "{{ provider }}"
    state: present
    partition: Common
    name: "{{ application }}_vs"
    description: "{{ scope + ', ' + source_allowlist }}"
    metadata: "{{ vs_metadata }}"
    destination: "{{ int_vs_ip }}"
    port: '443'
    pool: "{{ application }}_pool"
    snat: "{{ snat_pool_default }}"
    profiles: "{{ vs_config_map['profiles'] }}"
    policies: "{{ vs_config_map['policies'] | default([]) }}"
    security_log_profiles: "{{ vs_config_map['security_log_profiles'] }}"
    irules: "{{ (irules_list is defined and irules_list | length>0) | ternary(vs_config_map['irules'], []) }}"
    check_profiles: no
    default_persistence_profile: "{% if lb_mode is defined and lb_mode == 'lb' and ( pool_members_aggregate_list | length ) > 1 %}default_cookie_insert_persistence{% else %}{% endif %}"
  delegate_to: localhost
```

##### EXPECTED RESULTS

At every playbook run, the metadata should be set as defined in the task, consistently.
##### ACTUAL RESULTS

Looking in the /var/log/audit file of the target BIG-IP I can see that in every 1 of 2 runs the metadata is not set.

1st run:
```
[...]

PR #2513, opened 2026-02-18
fix: metadata issue where defaults are set when updating w/existing v…

joelkeener:
Fix for issue #2509 that prevents updating metadata with defaults when the user provided metadata is identical to the what is already configured on the virtual server.
Fixed a second issue where specifying only part of the existing metadata would not remove the parts that are missing from the new metadata.

### DIFF
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_virtual_server.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_virtual_server.py
@@ -3440,18 +3440,21 @@ class Difference(object):
 
     @property
     def metadata(self):
+        if not self.want.insert_metadata:
+            return None
         if self.want.metadata is None:
             return None
-        elif len(self.want.metadata) == 0 and self.have.metadata is None:
-            return None
-        elif len(self.want.metadata) == 0 and not self.want.insert_metadata:
-            return None
-        elif len(self.want.metadata) == 0 and self.want.insert_metadata:
+        if len(self.want.metadata) == 0 and self.want.insert_metadata:
             return []
-        elif self.have.metadata is None:
+        if self.want.insert_metadata != self.have.insert_metadata:
             return self.want.metadata
-        result = self._diff_complex_items(self.want.metadata, self.have.metadata)
-        return result
+        if self.have.metadata is None:
+            return self.want.metadata
+        w = self.to_tuple(self.want.metadata)
+        h = self.to_tuple(self.have.metadata)
+        if set(w) == set(h):
+            return None
+        return self.want.metadata
 
     @property
     def type(self):


tpitkinf5 (changes requested):
# Pull Request Review: Metadata Fix for Issue #2509

Thank you for working on this fix! I've reviewed the changes and while the commit does address the `issubset()` bug that was causing the alternating behavior, it unfortunately introduces several new issues that could impact metadata management.

## Issues Identified

### 1. **Missing Metadata Content Comparison**

**Problem:** The new code doesn't compare metadata content when both `want.insert_metadata` and `have.insert_metadata` are `True`.

**Scenario:**
```python
want.metadata = [{'name': 'foo', 'value': 'new_value'}]
have.metadata = [{'name': 'foo', 'value': 'old_value'}]
want.insert_metadata = True
have.insert_metadata = True
```

**Current behavior:** Returns `None` (no change detected)
**Expected behavior:** Should detect the value changed and return `want.metadata`

### 2. **Potential Crash: Missing None Check**

**Problem:** The code will crash if `want.metadata` is `None`:

```python
elif len(self.want.metadata) == 0 and self.want.insert_metadata:
    return []
```

**Error:** `TypeError: object of type 'NoneType' has no len()`

The old code handled this with an upfront check: `if self.want.metadata is None: return None`

### 3. **Behavior Change for Empty Metadata on New Resources** *(Minor - may be acceptable)*

**Note:** Both the current fix and my recommended fix share this behavior change.

**Scenario:**
```python
want.metadata = []
have.metadata = None  # New virtual server
want.insert_metadata = True
```

**Old behavior:** Returns `None` (no action)
**New behavior:** Returns `[]` (attempts to set empty metadata)

The old code had: `elif len(self.want.metadata) == 0 and self.have.metadata is None: return None`

**Impact:** This scenario is rare (user explicitly setting `metadata: []` on a new resource) and the behavior change is likely harmless - setting empty metadata on the device is typically a no-op. However, if you want to preserve the exact old behavior, add this check before the "Clear metadata" condition:

```python
# Preserve old behavior: don't set empty metadata on new resources
if len(self.want.metadata) == 0 and self.have.metadata is None:
    return None
```

### 4. **Fails to Detect Initial Metadata Creation**

**Problem:** When adding metadata to an existing virtual server that has no metadata.

**Scenario:**
```python
want.metadata = [{'name': 'foo', 'value': 'bar'}]
have.metadata = None  # No existing metadata
want.insert_metadata = True
have.insert_metadata = True  # Defaults to True
```

**Current behavior:** Returns `None` (no change)
**Expected behavior:** Should return `want.metadata` to create the metadata

The old code handled this: `elif self.have.metadata is None: return self.want.metadata`

## Recommended Fix

Here's a corrected version that addresses both the original bug and the new issues:

```python
@property
def metadata(self):
    # Don't manage metadata if insert_metadata is False
    if not self.want.insert_metadata:
        return None

    # Handle None case to prevent crashes
    if self.want.metadata is None:
        return None

    # Clear metadata if user provides empty list with insert_metadata=True
    if len(self.want.metadata) == 0 and self.want.insert_metadata:
        return []

    # Detect when insert_metadata flag changes
    if self.want.insert_metadata != self.have.insert_metadata:
        return self.want.metadata

    # Handle initial metadata creation
    if self.have.metadata is None:
        return self.want.metadata

    # Compare metadata content - FIX: Use equality instead of subset
    # This is the key change to fix issue #2509
    w = self.to_tuple(self.want.metadata)
    h = self.to_tuple(self.have.metadata)
    if set(w) == set(h):  # Changed from issubset to equality
        return None
    return self.want.metadata
```

## Why This Fix Works

The original bug was in `_diff_complex_items()`:
```python
if set(w).issubset(set(h)):  # BUG: subset check
    return None
```

The **subset check** caused the issue because user metadata is always a subset of device metadata (which includes ansible-managed metadata). By changing to **equality check** (`set(w) == set(h)`), we properly detect when user metadata differs from device metadata, regardless of the ansible-managed metadata fields.

## Summary

Your approach of simplifying the logic and avoiding `_diff_complex_items()` was correct, but the implementation needs to include:
1. ✅ Avoid the problematic subset logic (your fix does this)
2. ❌ Compare metadata content when both sides have `insert_metadata=True` (missing)
3. ❌ Handle `None` values safely (missing)
4. ❌ Preserve behavior for edge cases like initial creation (missing)

The recommended fix above addresses all scenarios while maintaining backward compatibility and fixing the original issue.

tpitkinf5 (approved):
Looks good, thanks!

Issue #2505, opened 2025-12-03
Cannot updated certifikate/priv-key used in SSL-Profile using bigip_ssl_certificate and bigip_ssl_key


Ichnafi:
##### COMPONENT NAME
- bigip_ssl_certificate
- bigip_ssl_key

(bigip_ssl_key_cert will not work as well, see #2504)

##### ANSIBLE VERSION
```
ansible [core 2.18.1]
  executable location = /opt/ansible/venv/bin/ansible
  python version = 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] (/opt/ansible/venv/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```
##### BIGIP VERSION
```
Sys::Version
Main Package
  Product     BIG-IP
  Version     17.5.1.3
  Build       0.0.19
  Edition     Point Release 3
  Date        Thu Oct  9 11:01:53 PDT 2025

```
##### CONFIGURATION
default

##### OS / ENVIRONMENT
Multiple different Linux Systems and Distributions.

##### SUMMARY
We need to replace certificate/priv-key that already in use in a client-SSL Profile. Since the module `bigip_ssl_key_cert` for replacing both at the same time dies not work for us (#2504) we tried to use `bigip_ssl_certificate` and `bigip_ssl_key` to replace ciertificate in one task and the priv-key in a second task.
If the certificate/priv-key are not used in any SSL-profile, this works as expected. If the certificate/priv-key are used in a SSL-profile, the first task will fail, because new certificate does not match the old priv-key. 

It's kind of a chicken-and-egg problem.

##### STEPS TO REPRODUCE
Run a the following tasks, to overwrite existing certificate/prov-key with new ones.

```
- name: Update Zertifikat {{ cert_name }}
  f5networks.f5_modules.bigip_ssl_certificate:
    name: "{{ cert_name }}"
    state: present
    content: "{{ lookup('file', cert_tempfile.path) }}"
    true_names: true
    provider: "{{ bigip_provider }}"
  register: result_update_cert_command
  delegate_to: localhost
  
- name: Update Private Key {{ cert_name }}
  when: result_update_cert_command.changed
  f5networks.f5_modules.bigip_ssl_key:
    name: "{{ cert_name }}"
    state: present
    content: "{{ lookup('file', key_tempfile.path) }}"
    passphrase: "{{ privkey_passphrase }}"
    true_names: true
    provider: "{{ bigip_provider }}"
[...]

Issue #2504, opened 2025-12-03
bigip_ssl_key_cert - Overwriting existing Certificates and Keys fails


Ichnafi:
##### COMPONENT NAME
bigip_ssl_key_cert

##### ANSIBLE VERSION
```
ansible [core 2.18.1]
  executable location = /opt/ansible/venv/bin/ansible
  python version = 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] (/opt/ansible/venv/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```
##### BIGIP VERSION
```
Sys::Version
Main Package
  Product     BIG-IP
  Version     17.5.1.3
  Build       0.0.19
  Edition     Point Release 3
  Date        Thu Oct  9 11:01:53 PDT 2025

```
##### CONFIGURATION
default

##### OS / ENVIRONMENT
Multiple different Linux Systems and Distributions. 

##### SUMMARY
module bigip_ssl_key_cert has error trying to replace existing certificate. (similar to #2058 , but kind of different).

##### STEPS TO REPRODUCE
Upload non existing certificate/priv-key using this task
```
- name: Update Zertifikat und Private-Key {{ cert_name }}
  f5networks.f5_modules.bigip_ssl_key_cert:
    key_content: "{{ lookup('file', key_tempfile.path) }}"
    key_name: cert2
    cert_content: "{{ lookup('file', cert_tempfile.path) }}"
    cert_name: cert2
    state: present
    true_names: true
    provider: "{{ bigip_provider }}"
  register: result_update_cert_command
  delegate_to: localhost
```
Rerun this task with same or different certificate/priv-key

##### EXPECTED RESULTS
- On rerun with same certificate/priv-key, ansible should show OK (no change).
- On rerun with different certificate/priv-key, ansible should show CHANGED.

Both tasks fail with the same error message

##### ACTUAL RESULTS
- Rerun with same certificate/priv-key fails with following error message.
```
TASK [Update Zertifikat und Private-Key test-zertifikat] *************************************************************************************************************************************
task path: /home/user/development/bamf-zertifikats-service-api/tasks/update_certificate.yml:37
Read vars_file 'vars/bigip-provider.yml'
Read vars_file 'vars/credentials-vault.yml'
File lookup using /tmp/ansible.ce4ekey9.key as file
File lookup using /tmp/ansible.5ghtlold.crt as file
[...]

PR #2515, opened 2026-05-18
fix: surface bigip error messages when ssl cert/key transaction fails

joelkeener:
**Problem**

`TransactionContextManager` in `icontrol.py` raises bare `Exception` with no message when transaction creation or commit fails. This makes it impossible to diagnose failures in any module that uses transactions (e.g., `bigip_ssl_key_cert`, `bigip_ssl_certificate`, `bigip_ssl_key`). Users see only a traceback ending at `icontrol.py` with `Exception` — the actual BIG-IP error message is swallowed.

Reported in #2504 and #2505.

**Fix**

Replaced both bare `raise Exception` statements in `TransactionContextManager` (`__enter__` and `__exit__`) with `F5ModuleError` that includes the actual BIG-IP response message, or falls back to HTTP status code and response body when JSON is unavailable.

**Testing**

- Added 6 unit tests to `test_bigip_ssl_key_cert.py` covering transaction creation failure, commit failure, success, and exception passthrough scenarios.

- Validated updating existing cert/key referenced by ssl profile on a BIG-IP device:

  - Updating cert/key with matching pair succeeds ✓

  - Updating cert only with mismatched key fails with clear error: `"Failed to commit transaction: ...key and certificate do not match"` ✓

  - Updating both cert and key atomically with a new matching pair succeeds ✓

### DIFF
--- a/ansible_collections/f5networks/f5_modules/plugins/module_utils/icontrol.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/module_utils/icontrol.py
@@ -270,7 +270,19 @@ class TransactionContextManager(object):
         )
         resp = self.client.api.post(uri, json={})
         if resp.status not in [200]:
-            raise Exception
+            try:
+                response = resp.json()
+                if 'message' in response:
+                    raise F5ModuleError(
+                        'Failed to create transaction: {0}'.format(response['message'])
+                    )
+            except ValueError:
+                pass
+            raise F5ModuleError(
+                'Failed to create transaction, status: {0}, contents: {1}'.format(
+                    resp.status, resp.content
+                )
+            )
         try:
             response = resp.json()
         except ValueError as ex:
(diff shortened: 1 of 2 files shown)


crosbygw (approved):
1. Minor cosmetic issue — icontrol.py:273-285 and 309-322
If the BIG-IP JSON response contains a 'message' key with a falsy value (e.g., {'message': ''} or {'message': None}), the error message will read "Failed to create transaction: " or "Failed to create transaction: None". Unlikely in practice but worth noting. No action required.
2. Behavioral change (intentional) — F5ModuleError replaces bare Exception
Callers already handle F5ModuleError at higher levels (Ansible's module runner catches it), so this is correct and consistent with the established pattern throughout icontrol.py.
3. Test coverage gap — header cleanup not asserted
The __exit__ method pops X-F5-REST-Coordination-Id from headers before the commit attempt, but none of the new tests (test_exit_skips_commit_on_exception, test_exit_failure_with_message, test_exit_failure_without_message) verify this header cleanup occurred. Not a bug, but a missed assertion opportunity.

joelkeener:
1. Unlikely in practice but worth noting. No action required.
2. This intentional change is correct and consistent with the established pattern throughout icontrol.py
3. There's no real coverage gap here. The header cleanup is implicitly validated by the tests already passing without KeyError. The headers.pop('X-F5-REST-Coordination-Id') on line 294 would raise a KeyError if the header wasn't there, and it runs before any of the code the new tests are exercising. The new tests don't need explicit header assertions.

PR #2518, opened 2026-06-10
fix: allow underscore in wideip fqdn

joelkeener:
The `is_valid_fqdn()` validation function in `module_utils/common.py` rejects DNS names containing underscores, causing `bigip_gtm_wide_ip` to fail with `"The provided name must be a valid FQDN"` for names like `my_app.example.com` or `_sip._tcp.example.com`.

Underscores are valid in DNS names and accepted by BIG-IP. They are also required by SRV, DMARC, and DKIM record types.

Added `_` to the `is_valid_fqdn` regex character class. The `is_valid_hostname()` function was intentionally left unchanged, as hostname validation correctly excludes underscores per RFC 952/1123.

**Affected module:** `bigip_gtm_wide_ip` — the only consumer of `is_valid_fqdn`

### DIFF
--- a/ansible_collections/f5networks/f5_modules/changelogs/fragments/fix-gtm-spaces-in-names.yaml
+++ b/ansible_collections/f5networks/f5_modules/changelogs/fragments/fix-gtm-spaces-in-names.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+    - Allow spaces in GTM/DNS object names by encoding them in REST API URI construction
--- a/ansible_collections/f5networks/f5_modules/changelogs/fragments/fix-gtm-wideip-underscore-fqdn.yaml
+++ b/ansible_collections/f5networks/f5_modules/changelogs/fragments/fix-gtm-wideip-underscore-fqdn.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+    - Allow underscores in GTM Wide IP names, which are valid in DNS names and the BIG-IP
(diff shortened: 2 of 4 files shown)


tpitkinf5 (approved):

crosbygw (approved):
Code Review: underscore-in-wideip-fqdn vs devel
1. Regex change in is_valid_fqdn — common.py:246
The regex change from [A-Z0-9-*] to [A-Z0-9_*-] is correct. Adding _ to the character class and moving - to the end is proper regex convention. The functional change (allowing underscores in FQDNs) is reasonable for BIG-IP GTM Wide IP names, where underscores are common (e.g., SRV records like _sip._tcp.example.com).
Minor note: The function's docstring still references RFC 3696 and RFC 1123 without noting this deviation. Consider updating it to mention that underscores are allowed as a BIG-IP-specific extension.
2. Changelog fragments
- Both changelog YAML files are missing trailing newlines. This won't cause functional issues but may trigger lint warnings in CI.
- fix-gtm-spaces-in-names.yaml describes "encoding spaces in REST API URI construction," but that change doesn't appear in this branch's diff — the space-encoding logic in transform_name (common.py:265) already exists on devel. Verify whether this fragment belongs in this branch or is an orphan from a separate merge.
3. No test coverage
There are no tests for is_valid_fqdn in the test suite. While not introduced by this change, regex modifications are error-prone. Adding unit tests covering normal FQDNs, underscore FQDNs, wildcard FQDNs, and invalid inputs would reduce regression risk.

Summary: The core code change is correct and low-risk. Actionable items:
1. Verify fix-gtm-spaces-in-names.yaml belongs in this branch
2. Add trailing newlines to the YAML changelog files
3. (Optional) Update the docstring to note the RFC deviation
4. (Optional) Add unit tests for is_valid_fqdn

PR #2519, opened 2026-06-15
docs: : correct force param for apm import

joelkeener:
MR description:

The force parameter description for the bigip_apm_policy_import module incorrectly stated that an existing policy would be overwritten by the new import. In reality, due to the nature of APM, the existing policy is not replaced in place. Instead, a new policy is created with a number appended to the previous name.

This change updates the documentation to accurately reflect the actual behavior of the module.


crosbygw (approved):
Looks good

PR #2520, opened 2026-06-24
feat: allow configuration of the expandSubcollections setting

joelkeener:
### Summary

Adds a new `expand_subcollections` boolean parameter (default: `true`) to the `bigip_device_info` module, allowing users to control whether the `expandSubcollections=true` query parameter is sent to the BIG-IP REST API. When disabled, the module fetches subcollection data via individual API calls instead of inline expansion, preserving the module's data contract while improving performance for large configurations.

### Problem

Several `bigip_device_info` subsets hardcode `expandSubcollections=true` in their API queries, which causes the BIG-IP device to return all sub-collection data inline with each resource. For large configurations, this significantly increases response size and query time. Users had no way to disable this behavior.

### Solution

- A new top-level module parameter `expand_subcollections` (type: `bool`, default: `true`) conditionally includes or omits the `expandSubcollections` query parameter across 16 API call sites in 12 FactManager classes.
- When `expand_subcollections: false`, the module makes separate API calls to fetch subcollection data (members, devices, virtual servers, rules, profiles, policies, interfaces) so that all documented return fields remain populated. This avoids a silent data regression where nested fields would appear empty even though children exist on the device.
- Default `true` preserves full backward compatibility.

### Affected Subsets

`device-groups`, `gtm-a-pools`, `gtm-aaaa-pools`, `gtm-cname-pools`, `gtm-mx-pools`, `gtm-naptr-pools`, `gtm-srv-pools`, `gtm-servers`, `ltm-pools`, `snat-pools`, `ltm-policies`, `virtual-servers`, `vlans`, `management-routes`

All other subsets are unaffected by this parameter.

### Changes

**`plugins/modules/bigip_device_info.py`**
- Added `expand_subcollections` to `DOCUMENTATION`, `EXAMPLES`, and `ArgumentSpec`
- Wrapped 16 API call sites across 12 FactManager classes with conditional query string logic
- Added `_read_subcollection()` helper methods to 9 managers (`DeviceGroupsFactManager`, `GtmServersFactManager`, `GtmAPoolsFactManager`, `GtmAaaaPoolsFactManager`, `GtmCnamePoolsFactManager`, `GtmMxPoolsFactManager`, `GtmNaptrPoolsFactManager`, `GtmSrvPoolsFactManager`, `LtmPolicyFactManager`, `VirtualServersFactManager`, `VlansFactManager`) to fetch child resources individually when `expand_subcollections=false`
- Updated `LtmPoolsFactManager.read_facts()` to use existing `read_member_from_device()` when subcollections are not expanded

**`tests/unit/.../test_bigip_device_info.py`**
- Added 7 unit tests covering:
  - Default parameter value (`True`)
  - Explicit `False` acceptance
  - Query string includes `expandSubcollections=true` when enabled
  - Query string excludes `expandSubcollections` when disabled
  - Correct behavior with `partition=all` (both `True` and `False`)
  - Parameter test coverage across `GtmServersFactManager`, `LtmPoolsFactManager`, and `VirtualServersFactManager`

**`changelogs/fragments/feat-bigip-device-info-expand-subcollections.yaml`**
- Changelog fragment for `minor_changes`

### Usage Example

```yaml
# Collect LTM pool info without inline subcollection expansion (faster for large configs)
- name: Collect LTM pool information
  bigip_device_info:
    gather_subset:
      - ltm-pools
    expand_subcollections: false
    provider:
      server: lb.mydomain.com
      user: admin
      password: secret
  delegate_to: localhost
```

### Backward Compatibility

✅ Fully backward compatible — `expand_subcollections` defaults to `true`, preserving existing behavior. When set to `false`, all documented return fields remain populated via individual subcollection API calls.


crosbygw (changes requested):
Code review complete for branch AUTOTOOL-5278-expand-subcols vs devel. This adds an expand_subcollections toggle to bigip_device_info across 14 fact managers.

**Issues to fix**
1. Incorrect version_added — bigip_device_info.py:54 says "1.28.0", but the collection is at 1.40.0-devel (latest release 1.39.0). 1.28.0 is already released, so this will fail the Ansible version_added sanity test. Should be "1.40.0".
2. Missing trailing newline — changelogs/fragments/feat-bigip-device-info-expand-subcollections.yaml lacks a trailing newline; yamllint will flag no-new-line-at-end-of-file.

**Non-blocking observations**

- _read_subcollection is copy-pasted into 11 manager classes; a shared BaseManager helper would cut maintenance surface (consistent with existing duplication, so a judgment call).
- _read_subcollection doesn't status-check responses, unlike siblings like read_member_from_device which raise F5ModuleError on non-200. A JSON-parseable 4xx/5xx body passes through silently and degrades to "empty subcollection," masking API failures.
- The "improves performance" claim may mislead for VirtualServersFactManager and GtmServersFactManager, where subcollection fetches run sequentially (not via ThreadPoolExecutor), trading one expanded query for N sequential round-trips.

**Verified correct**

- Query-string construction for both partition=all and filtered branches across all managers.
- Data shape from _read_subcollection matches inline expansion; downstream 'items' not in ... guards handle both paths.
- LtmPools default-true path preserved; new elif branch routes correctly.

joelkeener:
Applied both fixes:

1. **`version_added`** — updated to `"1.43.0"` (the next release version), good catch on the sanity test.
2. **Trailing newline** — added to the changelog fragment.

On the non-blocking items:

- **`_read_subcollection` duplication** — Agree a `BaseManager` helper would be cleaner, but since the existing codebase already follows this copy-per-manager pattern I'd rather not mix a refactor into a feature PR. Happy to do a follow-up cleanup if the team wants it.
- **Missing status check in `_read_subcollection`** — Fair point. I'll file a follow-up to add `F5ModuleError` on non-200 responses, consistent with `read_member_from_device`. Keeping it out of this PR to limit scope.
- **Performance claim** — You're right that for `VirtualServersFactManager` and `GtmServersFactManager` the tradeoff is different (N sequential calls vs one expanded query). The real win is for memory-constrained BIG-IPs where inline expansion causes API timeouts. I've clarified the description to say "can reduce memory pressure on the device" rather than blanket "improves performance."

Both fixes are pushed, tests still green (8/8).

crosbygw (approved):
Great, tahnks!

PR #2523, opened 2026-07-28
ci: redhat certification tests

joelkeener:
## Summary

Adds a local Red Hat Ansible Automation Hub certification check that replicates all three certification gates (build/import, lint, sanity) across the full ansible-core matrix. Fixes a root-cause certification failure that caused rejection of v1.39.0 by removing two orphan action plugins that had no matching modules and were suppressed by stale sanity ignore files through every ansible-core version.

## Changes

- Certification workflow

certification.yml — scoped to PRs targeting devel and merge queue entries so certification failures block merges
Local certification tooling

- Skill
  - .github/skills/redhat-cert-test/scripts/redhat-cert-test.sh
    - Add a standalone, CI-equivalent `ansible-lint --profile=production` step (installs resolved `ansible-lint` version and runs from the collection root).
    - Disable galaxy-importer internal linting (`RUN_ANSIBLE_LINT=False`) so lint runs exactly once and with the same flags Red Hat uses.

- Changelogs
  - ansible_collections/f5networks/f5_modules/changelogs/changelog.yaml
  - ansible_collections/f5networks/f5_modules/changelogs/fragments/feat-bigip-device-info-expand-subcollections.yaml
    - Apply formatting fixes (YAML indentation and long-line wrapping) to satisfy `ansible-lint --profile=production`.

- Collection fixes and cleanup
  - ansible_collections/f5networks/f5_modules/plugins/module_utils/common.py
  - ansible_collections/f5networks/f5_modules/plugins/module_utils/ipaddress.py
  - ansible_collections/f5networks/f5_modules/plugins/modules/bigip_pool.py
  - ansible_collections/f5networks/f5_modules/plugins/modules/bigip_pool_member.py
  - ansible_collections/f5networks/f5_modules/plugins/modules/bigip_gtm_pool_member.py
  - ansible_collections/f5networks/f5_modules/plugins/modules/bigip_command.py
  - ansible_collections/f5networks/f5_modules/plugins/modules/bigip_imish_config.py
  - ansible_collections/f5networks/f5_modules/plugins/action/* (removed orphaned action plugins)
  - ansible_collections/f5networks/f5_modules/tests/sanity/ignore-*.txt (stale ignore entries removed)
    - Guard imports, add fallbacks, and remove stale action-plugin ignores found by sanity/import checks.

## Testing

- Manual / local verification:
  - Ran `ansible-lint --profile=production` (ansible-lint 24.12.2) from `ansible_collections/f5networks/f5_modules` — passed after changelog fixes.
  - Ran the updated skill:
    - Standalone lint step executed and passed.
    - Full `ansible-test sanity` matrix run (ansible-core 2.16, 2.18, 2.20) in Docker (CI-identical) — passed.
  - Saved logs: lint output and skill run logs available in the artifacts directory under `.github/skills/redhat-cert-test/artifacts/` for inspection.

## Checklist

- [x] Sanity matrix (2.16 / 2.18 / 2.20) passes (Docker CI-equivalent)
- [x] `ansible-lint --profile=production` passes against collection root


crosbygw (changes requested):
Code Review: AUTOTOOL-5313-redhat-cert vs devel
Bugs
1. remove_default_spec fallback has wrong semantics (medium) — plugins/modules/bigip_pool.py:604, bigip_pool_member.py:572, bigip_gtm_pool_member.py:406
Fallback just returns argument_spec unchanged. Real netcommon impl mutates the spec, deleting "default" keys. Callers use it as a statement (e.g. bigip_pool.py:1483), relying on the side effect. When netcommon is absent, defaults leak into aggregate sub-specs and get re-applied per list element. Fix:
def remove_default_spec(spec):
    for item in spec:
        if "default" in spec[item]:
            del spec[item]["default"]
2. validate_ip_address fallback accepts different inputs (low–medium) — plugins/module_utils/ipaddress.py:16
Fallback uses ip_interface(), which accepts "10.0.0.1/24"; netcommon's socket.inet_aton-based version does not. Divergent validation when the fallback is used.
3. galaxy.yml netcommon pin is likely too strict (medium)
ansible.netcommon: "8.6.1"
Bare version is treated as exact-match by ansible-galaxy. Blocks installs where a different netcommon is already resolved. Use ">=8.6.1" if the intent is a minimum.
4. Hard dependency vs import guards contradiction
galaxy.yml still declares netcommon as required, so the ImportError fallbacks are only reachable during sanity/import scans that don't install collection deps. Worth a one-line comment above one guarded import explaining this so future maintainers don't strip them.
Minor
- Action plugin fallback (bigip_imish_config.py:31): only raises on .run(), not __init__ — intentional and fine.
- Deletion of bigip.py/bigiq.py action plugins and matching ignore-2.13..2.18.txt entries is consistent; no dangling refs.
- .gitignore addition is fine.
Bottom line
Real issues to address: #1, #2, #3. #4 is a documentation nit. The action plugin cleanup is coherent.

joelkeener:
1) remove_default_spec:
- Implemented fallback that mutates the argument_spec in-place and removes "default" keys (recurses into nested option dicts), matching ansible.netcommon semantics. Changed: bigip_pool.py, bigip_pool_member.py, bigip_gtm_pool_member.py.

2) validate_ip_address:
- Replaced ip_interface-based fallback with validators that reject CIDR-suffixed inputs and validate plain IPv4/IPv6 addresses (IPv4Address/IPv6Address), aligning behavior with netcommon. Changed: plugins/module_utils/ipaddress.py.

3) galaxy.yml netcommon pin:
- Loosened dependency from "8.6.1" to ">=8.6.1" to express a minimum version rather than an exact match. Changed: ansible_collections/f5networks/f5_modules/galaxy.yml.

4) Hard-dependency vs import guards:
- Added a one-line comment above the guarded import explaining the try/except is intentional for sanity/import scans so maintainers don’t remove the fallback. (Comment added alongside the remove_default_spec guarded import.)

crosbygw (approved):
Changes look good!

PR #2528, opened 2026-08-25
feat: add ipport as valid cmp_hash choice [AUTOTOOL-5333]

joelkeener:
### Summary

Adds support for the `ipport` CMP hash disaggregation method in the `bigip_vlan` module, resolving a gap between module capabilities and BIG-IP device functionality. This choice was available on BIG-IP systems but missing from the module's documented options.

### Changes

**Module:**
- Added `ipport` to the `cmp_hash` parameter choices in `bigip_vlan.py`

**Tests:**
- Added `test_create_vlan_with_cmp_hash_ipport()` test case in `test_bigip_vlan.py` to verify the new choice is accepted and applied correctly

### Testing

- New unit test case validates VLAN creation with `cmp_hash='ipport'` parameter
- Test verifies module state change and parameter value assertion
- Existing test suite remains unaffected; all test patterns follow established conventions

### Checklist

- [x] Unit tests added and verified

### DIFF
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_vlan.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_vlan.py
@@ -74,8 +74,8 @@ options:
       - Specifies how the traffic on the VLAN is disaggregated. The value
         you select determines the traffic disaggregation method. You can choose to
         disaggregate traffic based on C(source-address) (the source IP address),
-        C(destination-address) (destination IP address), or C(default), which
-        specifies the default CMP hash uses L4 ports.
+        C(destination-address) (destination IP address), C(ipport) (source IP address
+        and port), or C(default), which specifies the default CMP hash uses L4 ports.
       - When creating a new VLAN, if this parameter is not specified, the default
         is C(default).
     type: str
--- a/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_vlan.py
+++ b/ansible_collections/f5networks/f5_modules/plugins/modules/bigip_vlan.py
@@ -90,6 +90,7 @@ options:
       - source
       - dst
       - src
+      - ipport
   dag_tunnel:
     description:
       - Specifies how the disaggregator (DAG) distributes received tunnel-encapsulated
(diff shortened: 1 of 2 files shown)


crosbygw (changes requested):
Code Review Summary: bigip_vlan.py — Add ipport choice for cmp_hash

Bug found:
- Duplicate test method at test_bigip_vlan.py:340 and test_bigip_vlan.py:369 — both named test_create_vlan_with_cmp_hash_ipport with identical bodies. Python silently overwrites the first with the second, making one copy dead code. Likely a copy-paste/merge artifact. Remove the duplicate.

Minor issues:
- Doc inconsistency: The choices: YAML block (bigip_vlan.py:93) now includes ipport, but the prose description (bigip_vlan.py:72-81) still only explains source-address, destination-address, and default. Add a sentence describing ipport for consistency.
- No synonym aliases for ipport: src-ip and dst-ip each accept 4 synonym spellings (e.g., src, source-address, source), but ipport only accepts the literal ipport — no alias like ip-port. Not a bug, just less flexible than the existing pattern; worth confirming this is intentional.

Verified correct:
- The new mapping in ModuleParameters.cmp_hash (bigip_vlan.py:509-510) correctly passes ipport through unmodified, and this matches the F5 tmsh cmp-hash accepted values (default | dst-ip | src-ip | ipport).
The main actionable item is the duplicate test method — everything else is minor polish.

joelkeener:
I did not add aliases for ipport as that is the name of the cmp-hash mode and it isn't a src|dst ip, but represents a mode that uses the F5 4 tuple (ingress src-ip:port/dst-ip:port and egress src-ip:port/dst-ip:port).

I did make these changes:
1. Duplicate test method removed — The second test_create_vlan_with_cmp_hash_ipport (lines 369-398) was a copy-paste artifact and has been removed. Only the first definition now exists.

2. Documentation updated — Added ipport to the prose description in the cmp_hash parameter doc so it now explicitly mentions all supported disaggregation methods: source-address, destination-address, ipport, and default.

crosbygw (approved):
Changes look good!

PR #2534, opened 2026-09-16
fix: gtm topology region to allow spaced names

joelkeener:
## Summary

Supports spaces in GTM topology region names and member values within the `bigip_gtm_topology_region` module. Region names containing spaces are now properly escape-quoted for tmsh execution via the bash endpoint, preventing bash from splitting names into separate arguments.

### Module
- `ansible_collections/f5networks/f5_modules/plugins/modules/bigip_gtm_topology_region.py`:
  - Added `tmsh_name` property to `ModuleManager` to ensure fully-qualified region names containing spaces are escape-quoted (`\"/Common/Test Space\"`) when formatted into tmsh bash payloads.
  - Updated `create_on_device()` and `update_on_device()` to use `self.tmsh_name`.
  - Added `name` property to `ModuleParameters` to strip redundant surrounding quotes from user input, ensuring consistent resource resolution across tmsh and REST endpoints.
  - Updated `UsableChanges.escape_spaces()` to handle negated member entries with spaces (`not state ...`, `not datacenter ...`).

### Tests
- `ansible_collections/f5networks/f5_modules/tests/unit/modules/network/f5/test_bigip_gtm_topology_region.py`:
  - Added `test_tmsh_name_with_spaces` to verify region names with spaces are properly escape-quoted.
  - Added `test_tmsh_name_without_spaces` to verify region names without spaces remain unquoted.
  - Added `test_tmsh_name_with_custom_partition_and_spaces` to verify partition preservation with spaces.
  - Added `test_module_parameters_strips_external_quotes_from_name` and `test_module_parameters_name_none`.
  - Added `test_escape_spaces_negate_state_with_spaces` and `test_escape_spaces_negate_datacenter_with_spaces`.
  - Added `test_create_on_device_with_spaces_in_name` and `test_update_on_device_with_spaces_in_name` to validate the generated bash API payload.

### Changelog
- `ansible_collections/f5networks/f5_modules/changelogs/fragments/fix-gtm-spaces-in-names.yaml`:
  - Added changelog fragment noting support for spaces in region names and member values.

## Testing

- **Unit tests:** Executed `pytest` against `test_bigip_gtm_topology_region.py` (24 passed).
- **Sanity checks:** Executed `ansible-test sanity` across modified modules, tests, and fragments (25 sanity checks passed).
- **Linting:** Validated changelog fragment with `ansible-lint` (0 failures, 0 warnings).

## Checklist

- [x] Unit tests pass
- [x] Ansible sanity and lint checks pass
- [x] Changelog fragment added

Ref:AUTOTOOL-5338


andreykashcheev (approved):
Looks good!

crosbygw (approved):
I reviewed the diff against devel for bigip_gtm_topology_region.py. This change extends an existing fix (spaces-in-values escaping for region members) to also cover spaces in the region name itself.

Result: No bugs found. Key points verified:
- ModuleParameters.name correctly strips surrounding quotes and is only read from want (never diffed against have), so no spurious "changed" state.
- escape_spaces's new not -prefix handling correctly isolates the prefix before matching against keys_with_potential_spaces; no ambiguity since none of those keys are substrings of each other.
- tmsh_name output format is consistent with existing escaping conventions elsewhere in the codebase (e.g. bigip_gtm_monitor_external.py).
- The resulting tmsh command payload is well-formed.

Minor (non-blocking) notes:
- Changelog fragment wording now aligns better with the module-prefixed naming convention used in other fragments — a slight improvement.
- New unit tests directly exercise the new behavior rather than just re-asserting old coverage.

One caveat: the pre-existing test suite has an unrelated failure (test_create_topology_region) tied to an Ansible version/environment issue — this fails identically on devel, so it's not caused by this change.

Overall: clean, well-tested follow-up applying the same escaping pattern to region names. No required changes.

PR #2535, opened 2026-09-16
fix: longer tmsh terminal prompts

joelkeener:
## Summary

Updates the BIG-IP terminal plugin to recognize longer tmsh prompts containing numerous parenthesized status indicators (e.g., MCP, AVC, config-sync, HA state) and newline-wrapped prompt characters that exceeded previous regex limits.

## Changes

- **Plugins**:
  - `ansible_collections/f5networks/f5_modules/plugins/terminal/bigip.py`: Added a regex pattern to `terminal_stdout_re` matching one or more parenthesized status segments followed by `>` or `#`, allowing long and wrapped prompts to match properly.
- **Tests**:
  - `ansible_collections/f5networks/f5_modules/tests/unit/plugins/terminal/test_bigip.py`: Added `test_tmos_prompt_many_parentheses_and_newline_hash` to verify `on_open_shell()` command execution with long prompts, and added `TestTerminalStdoutRegex` test cases validating positive and negative prompt matching.

## Testing

- Unit tests added and verified using `pytest`:
  - `ansible_collections/f5networks/f5_modules/tests/unit/plugins/terminal/test_bigip.py` (10/10 tests passed)

## Checklist

- [x] Unit tests pass

Ref:AUTOTOOL-5392


crosbygw (approved):
Review Summary: 686cc93c (fix: longer tmsh terminal prompts) vs devel

No functional bugs found. The change adds a new regex as the first entry in terminal_stdout_re to handle BIG-IP tmos prompts with more than 5 parenthesized segments. The old regex capped at {,5} groups, so prompts with 6+ groups fell back to a narrow match (\n# ) missing tmos, breaking downstream prompt-detection in netcommon's network_cli plugin.

Verified:
- The new pattern is a strict superset — matches identically to the old one on all existing ≤5-group prompts (no regression).
- Correctly handles the reported >5-group case.
- No practical regex-performance risk since network_cli bounds input to ~256-byte windows before matching.
- New unit tests reproduce the exact bug-report prompt and pass.
Minor, non-blocking observation: No changelog fragment was added under changelogs/fragments/, unlike the related prior fix (f385760c) and most recent bugfix commits in this repo. Easy to add if required for release notes.

PR #2536, opened 2026-09-17
fix: gather_subsets for gtm server vs stats 404 if name has space

joelkeener:
## Summary

Fixes an issue in `bigip_device_info` where gathering `gtm-servers` facts fails with a 404 error when a virtual server reference has spaces in its name. The module now gracefully handles missing virtual server statistics for spaced names instead of aborting fact collection.

### Module
- `ansible_collections/f5networks/f5_modules/plugins/modules/bigip_device_info.py`: In `GtmServersParameters._read_virtual_stats_from_device`, ignore 404 responses when querying `/stats` for virtual servers whose URI contains spaces or encoded spaces (`%20`), returning an empty dict for stats instead of raising `F5ModuleError`.

### Tests
- `ansible_collections/f5networks/f5_modules/tests/unit/modules/network/f5/test_bigip_device_info.py`: Added unit test cases covering 404 responses with spaces, 404 with `%20`, 404 without spaces (still raising `F5ModuleError`), 500 error handling, and 200 success responses.

### Documentation
- `ansible_collections/f5networks/f5_modules/changelogs/fragments/fix-gtm-servers-virtual-server-spaces.yaml`: Added changelog fragment documenting the bugfix.

## API Endpoints

| Method | Endpoint | Description |
|--------|----------|-------------|
| GET | `/mgmt/tm/gtm/server/<server>/virtual-servers/<virtual_server>/stats` | Virtual server statistics endpoint on BIG-IP; 404s are now ignored for spaced names |

## Testing

- Added 5 new unit tests in `test_bigip_device_info.py` covering all edge cases (spaced names, `%20` encoded names, standard names, error statuses).
- Verified all 13 unit tests in `test_bigip_device_info.py` pass.
- Verified manually using the `jdk_play.yml` playbook against a live BIG-IP with spaced GTM virtual server objects.

## Checklist

- [x] Unit tests pass
- [x] Documentation updated (changelog fragment added)

Ref:AUTOTOOL-5392


crosbygw (changes requested):
Code Review Summary — AUTOTOOL-5391-spaced-vs-name vs devel

Change: In bigip_device_info.py, GtmServersParameters._read_virtual_stats_from_device now suppresses a 404 (returns {}) instead of raising F5ModuleError, when the request URI contains a whitespace/%20 and ends in /stats. Intent: avoid failures when a GTM server's virtual-server name has a space.

**Issues found**
1. Moderate — regex is too broad (bigip_device_info.py:11017)
if (resp.status == 404 or response.get('code') == 404) and re.match(r'(.*(\s|(%20))+.*)/stats$', uri.rstrip()):
    return {}
It matches whitespace/%20 anywhere in the URI, not just in the virtual-server name segment. So a 404 gets silently swallowed even when the space is in the partition/server name and the virtual server itself was genuinely deleted/renamed (e.g. .../gtm/server/~My Partition~server1/virtual-servers/vs1/stats). This masks legitimate "not found" errors, not just the specific space-encoding bug being targeted. Suggest scoping the match to just the path segment immediately before /stats.
2. Minor — silent data loss (bigip_device_info.py:11017-11018)
When suppressed, {} is returned with no warning, so downstream stats fields (status, state, bits_per_sec_in, etc.) are silently omitted for that virtual server. Codebase precedent (e.g. bigip_command.py:461) uses self.module.warn(...) in similar cases — worth adding here for visibility.
3. Minor nit — style inconsistency, not a bug. New check uses response.get('code') vs. the existing pattern above using 'code' in response and response['code']. Functionally fine, just inconsistent style.
4. Test gap: New unit tests (test_bigip_device_info.py:303-411) cover space, encoded-space, non-space 404, 500 error, and success cases — but don't cover the false-positive scenario in issue #1 (space elsewhere in path, VS legitimately missing), which is the actual gap.

**Bottom line**

The fix works for the intended case but is broader than necessary, risking silently hidden genuine 404 errors when whitespace appears elsewhere in the URL path (e.g., partition names). Recommend narrowing the regex to only match the virtual-server name segment before merging.

joelkeener:
1. I have modified the regexp to only match when a uri ends with /stats, and has virtual-servers/ before a string that contains at least one space that is surrounded by any characters (the space can't be at the beginning or end, but any number of spaces can be present).

2. This warning will now be logged when the GTM Server (host.domain.com - CA) has a VS with a space in the name (host_domain_com - CA_HTTPS):
```
[WARNING]: Failed to gather Stats for Virtual Server because the name has a space. 
    This is a limitation of the iControl REST API endpoint: 
    https://10.145.38.86:443/mgmt/tm/gtm/server/~Common~host.domain.com%20-%20CA/virtual-servers/host_domain_com%20-%20CA_HTTPS/stats
```

3. Fixed

4. Added tests.

crosbygw (approved):
Changes look good

